PR #11, opened 2020-10-02
Make Emboss work under Python 2 again.

reventlov:
Temporary measure for customers who are still using Emboss under Python 2.

Also fix a couple of minor lint errors introduced by the same pull request that broke Python 2.


robrussell (approved):

AaronWebster:
Thanks Rob.

PR #18, opened 2021-09-20
Compatibility improvements for Arduino ESP32/ESP8266 platforms.

AaronWebster:
- Renames generated .inc files to .h as these are not recognized by the Arduino IDE.
- Adds minimum keyword definitions to reserved word list.

Test cases build successfully under Arduino 1.18.5, though there are a few path modifications that had to be made.  These translations will be applied via external means if/when an official Arduino library is released.


reventlov:
The changes to `reserved_words` are backwards-incompatible, and thus should not be merged.  There are already a _lot_ of `#define`s in various header files that we don't ban.

AaronWebster:
Done.

PR #20, opened 2021-10-12
Move generate_grammar_md to data deps in docs_are_up_to_date_test

AaronWebster:
generate_grammar_md is a py_binary and should be listed in data instead of deps.

### DIFF
--- a/compiler/front_end/BUILD
+++ b/compiler/front_end/BUILD
@@ -452,10 +452,9 @@ py_test(
     name = "docs_are_up_to_date_test",
     srcs = ["docs_are_up_to_date_test.py"],
     data = [
+        ":generate_grammar_md",
         "//doc:grammar_md",
     ],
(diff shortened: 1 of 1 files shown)


reventlov:
I think you're just trading a quasi-supported Bazel feature (`import` from a `py_binary` in `deps`) for an unsupported Bazel feature (`import` from a `py_binary` in `data`).

`docs_are_up_to_date_test` is importing `generate_grammar_md.py`, not invoking the `generate_grammar_md` binary.  That's fine from a Python standpoint, but Bazel doesn't really support the notion of having a single `.py` file that can be either a binary's entry point or an importable module.

I think the proper fix is to move the guts of `generate_grammar_md.py` into a new `.py`, add a `py_library` rule for *that* new `.py`, and then have `docs_are_up_to_date_test` and `generate_grammar_md` use the new library.

PR #36, opened 2022-04-05
Implement the `$next` keyword for convenient packed structures.

reventlov:
This change adds a new keyword, `$next`, which may be used in the offset
expression of a physical field, and which translates to "the offset of
the first byte after the end of the syntactically previous physical
field."

Essentially, it allows `struct`s or `bits` like:

    struct Foo:
      0     [+4]  UInt  a
      $next [+2]  UInt  b
      $next [+1]  UInt  c
      $next [+4]  UInt  d

... where each field is packed directly after the previous one.  See the
updated documentation in `doc/language-reference.md` for details.

This also contains a few incidental changes:

*   A new function in `ir_util`, for scanning subtrees of the IR.  Note
    that a similar function used to exist; this is really a resurrection
    of that functionality.
*   `synthetics.synthesize_fields` has been renamed to
    `synthetics.desugar`, since it now does more than just synthesize
    fields.
*   The "how to use the Emboss compiler" sections of `guide.md` have been
    expanded.
*   The comment in `compiler/front_end/docs_are_up_to_date_test.py` has
    been updated with the correct paths for the open-source Emboss.


compiler/util/traverse_ir.py
```diff
@@ -281,3 +281,31 @@ def do_something(expression, field_name=None):
...
+  """Traverse a subtree of an IR, executing the given actions.
+
+  fast_traverse_node_top_down is like fast_traverse_ir_top_down, except that:
+
+  It may be called on a subtree, instead of the top of the IR.
+
+  It does not have any built-in incidental actions.
+
+  Arguments:
+    ir: An ir_pb2.Ir object to walk.
```
robrussell, line 297: Should be `node`. Same in skip_descendants_of description.

compiler/front_end/synthetics.py
```diff
@@ -215,23 +216,118 @@ def _add_size_virtuals(structure, type_definition):
...
   _add_size_virtuals(structure, type_definition)
   _add_size_bound_virtuals(structure, type_definition)
 
 
-def synthesize_fields(ir):
-  """Adds synthetic fields to all structures.
+def desugar(ir):
+  """Translates pure syntactic sugar to its desugared form.
+
+  Replaces `$then` symbols with the start+length of the previous physical
```
robrussell, line 314: Should this be $next instead of $then?

compiler/front_end/synthetics.py
```diff
@@ -215,23 +216,118 @@ def _add_size_virtuals(structure, type_definition):
...
+                  "`$next` may not be used in the first physical field of a " +
+                  "structure; perhaps you meant `0`?")
+    ])
+    return
+  original_location = expression.source_location
+  expression.CopyFrom(_NEXT_KEYWORD_REPLACEMENT_EXPRESSION)
+  expression.function.args[0].CopyFrom(last_location.start)
+  expression.function.args[1].CopyFrom(last_location.size)
+  expression.source_location.CopyFrom(original_location)
+  # DO NOT SUBMIT: Error messages for circular references?
```
robrussell, line 242: Anything left to do for this `DO NOT SUBMIT`?

compiler/front_end/dependency_checker.py
```diff
@@ -207,9 +224,10 @@ def strong_connect(node):
 
 def _find_object_dependency_cycles(ir):
   """Finds dependency cycles in types in the ir."""
-  dependencies = _find_dependencies(ir)
+  dependencies, errors = _find_dependencies(ir)
+  if errors:
+    return errors
   cycles = _find_cycles(dict(dependencies))
-  errors = []
```
robrussell, line 212: I think this means that `_find_object_dependency_cycles()` can now return `None` instead of `[]`. That should cause an exception later on when this append is attempted:
`errors + _find_object_dependency_cycles(ir)`
If _find_object_dependency_cycles(ir) returns a list then this becomes
`None + []`
which I suspect doesn't work.

compiler/back_end/cpp/testcode/next_keyword_test.cc
```diff
@@ -0,0 +1,55 @@
...
+#include <vector>
+
+#include "gtest/gtest.h"
+#include "testdata/next_keyword.emb.h"
+
+namespace emboss {
+namespace test {
+namespace {
+
+// [$default byte_order: "LittleEndian"]
```
robrussell, line 31: If this is meant to be kept then a short description would help. It looks like the contents of testdata/next_keyword.emb though?

robrussell (commented):
Nice feature. Just a couple questions & nits.

compiler/util/traverse_ir.py
reventlov: Fixed.  Apparently I am far too used to leaning on internal presubmits.

compiler/front_end/synthetics.py
reventlov: Fixed.

robrussell (approved):

compiler/front_end/synthetics.py
reventlov: Yes... the message is a bit less clear than I would prefer, but dependency cycles that include `$next` are correctly flagged as errors.

compiler/front_end/dependency_checker.py
reventlov: It was actually assigned a couple lines above.  I have refactored slightly to avoid reusing the variable and make things a bit clearer.

compiler/back_end/cpp/testcode/next_keyword_test.cc
reventlov: Added a small bit of context.

PR #52, opened 2022-08-17
Explicitly downcast in emboss_text_util.h

acsaeed:
The line in question generates a compilation failure in Fuchsia, "error: implicit conversion loses integer precision: int to ::std::uint8_t" due to the "-Wimplicit-int-conversion" diagnostic flag. Adding an explicit `static_cast` here allows compilation.

### DIFF
--- a/runtime/cpp/emboss_text_util.h
+++ b/runtime/cpp/emboss_text_util.h
@@ -69,7 +69,7 @@ class TextOutputOptions final {
 
   TextOutputOptions WithNumericBase(int new_value) const {
     TextOutputOptions result = *this;
-    result.numeric_base_ = static_cast<::std::uint8_t>(new_value);
+    result.numeric_base_ = static_cast</**/ ::std::uint8_t>(new_value);
     return result;
   }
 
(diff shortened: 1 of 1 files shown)


AaronWebster (commented):
LGTM, but drive-by question: Is there any reason that WithNumericBase uses int instead of uint8_t as an argument?

runtime/cpp/emboss_text_util.h
```diff
@@ -69,7 +69,7 @@ class TextOutputOptions final {
 
   TextOutputOptions WithNumericBase(int new_value) const {
     TextOutputOptions result = *this;
-    result.numeric_base_ = new_value;
+    result.numeric_base_ = static_cast<::std::uint8_t>(new_value);
```
reventlov, line 72: Please add /**/ between < and : to avoid the "<:" digraph that is enabled on some compilers; i.e.:

static_cast</**/ ::std::unit8_t>

acsaeed: Done

reventlov (approved):

PR #82, opened 2023-04-28
Design Sketch for String Support

reventlov:


doc/design_docs/strings.md
```diff
@@ -0,0 +1,321 @@
...
+
+    `CouldWriteValue()` should return `true` if the value is no longer than the
+    field and the value does not *contain* any instances of the terminator
+    byte.
+
+3.  `PaddedString`, a padded string.  A `PaddedString` with no arguments uses
+    space (`' '`, 32) as the padding value.  An optional argument can be used to
+    specify the padding -- a `PaddedString(0)`, for example, would be padded
+    with null bytes.  When reading, the end of the string is discovered by
+    walking *backwards* from the end until a non-padding byte is found, then
```
robrussell, line 89: It looks like strings are all LTR, should note somewhere that RTL is not yet supported.

doc/design_docs/strings.md
```diff
@@ -0,0 +1,321 @@
...
+    *   `\?` => 63 (part of the C standard, but rarely used)
+    *   `\\` => 92
+    *   <code>\x*hh*</code> => 0x*hh*
+
+    The following non-C-standard escapes should be allowed:
+
+    *   `\e` => 27 (not actually standard, but common)
+    *   <code>\d*nnn*</code> => *nnn*
+    *   <code>\x{*hh*}</code> => 0x*hh*
+    *   <code>\d{*nnn*}</code> => *nnn*
```
robrussell, line 132: Do the brace-delimited forms explicitly require 2 or 3 characters? If multibyte characters are allowed in the future then presumably there'd be some kind of syntax like \x{hhhh} for a two-byte value since \xhhhh would be ambiguous.

Even with single byte values, I think \x1 and \d10 are not not valid and should instead be \x01 and \d010. But are \x{1} and \d{10} valid?

doc/design_docs/strings.md
```diff
@@ -0,0 +1,321 @@
...
+
+
+#### String Comparisons
+
+Comparison operations (`==`, `<`, `>`, `>=`, `<=`, `!=`) should be allowed,
+since these can be handled by passing references to existing memory.
+
+Equality and inequality (`==` and `!=`) should be defined in the expected way:
+two strings are equal iff they are the same length and the corresponding bytes
+in each string have the same value, and they are unequal if they are not equal.
```
robrussell, line 284: Is it possible to compare strings of different types? Comparing a PaddedString and a ZString seems like it could take some thought.

robrussell (approved):

doc/design_docs/strings.md
```diff
@@ -56,37 +56,37 @@ important for now.  This decision can be revisited later.
...
-2.  `ZString` represents a terminated string.  A `ZString` with no arguments
-    uses a null byte (`'\0'`) as the terminator.  An optional argument can be
-    used to specify the terminator -- a `ZString(36)`, for example, would be
-    terminated by `$`.  When reading, the value returned is all bytes up to, but
-    not including, the first terminator byte.  When writing, for compatibility,
-    the entire field should be written, using the terminator value for padding
-    if there is extra space.  A second optional parameter can be used to specify
+2.  `ZString`, a terminated string.  A `ZString` with no arguments uses a null
+    byte (`'\0'`) as the terminator.  An optional argument can be used to
+    specify the terminator -- a `ZString(36)`, for example, would be terminated
```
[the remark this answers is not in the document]
reventlov: As far as I have seen, for narrow (ASCII, UTF-8, Latin-1) strings, yes.  I think the terminator is `'\0'` in 99.9% of cases, as well -- IIRC, the only places I've seen non-`'\0'` terminators were a couple of syscalls on MS-DOS and CP/M, which used `'$'`.

doc/design_docs/strings.md
reventlov: Unless I'm totally mistaken, LTR and RTL are handled out at the presentation layer -- Emboss does not really support either one, because that level of interpretation happens much later.

doc/design_docs/strings.md
reventlov: Updated the section on brace-delimited forms to be clearer.

Added text that `\x1` or `\d10` should be disallowed.  I'm not 100% convinced that it's the right choice, but a basic design principle I've always followed in Emboss is that it's much much easier to start strict and gradually allow more things than to start loose and try to disallow things after the fact.

I also added some text to the effect that unknown escape sequences should not be allowed.

doc/design_docs/strings.md
reventlov: I added some tentative implementation notes.

Basically, yes, once a string value is in the expression system, it's just a string, no matter what its physical representation is.  This is conceptually the same as comparing a value stored as `UInt` with one stored as `Bcd` -- the physical representation differs, but the semantic values are comparable.

PR #83, opened 2023-05-11
Add design for alternate enum cases

jasongraffius:


doc/design_docs/alternate_enum_cases.md
```diff
@@ -0,0 +1,323 @@
...
+This alternative design would still use `enum_case` or something similar, but
+not allow multiple case options to be asserted. Instead, either a new
+transition-specific case or a transitional attribute would be used to mark a
+transition in progress. For example:
+
+```
+enum Foo:
+  BAR             = 1  [(cpp) enum_case: "kCamelCase-transitional"]
+  BAZ             = 2  [(cpp) enum_case: "kCamelCase-transitional"]
+  MULTI_WORD_ENUM = 4  [(cpp) enum_case: "kCamelCase-transitional"]
```
reventlov, line 243: This seems easy to combine with your preferred design, by putting a `-deprecated` tag on the deprecated option(s) instead of a `-transitional` tag on the "not deprecated" option:

```
enum Foo:
  BAR_BAZ = 1
    [(cpp) enum_case: "SHOUTY_CASE -deprecated, CamelCase -deprecated, kCamelCase"]
```

(This would not need to be part of the initial implementation, so you could leave it as a "future work" item.)

reventlov (changes requested):
Overall looks good, just a couple of things to add to a "future work" section.

josh-conner:
I don't think of the constant syntax as an attribute of the data itself, but as an attribute of the environment where it will be used. Would it be possible to implement this as a command-line option?

doc/design_docs/alternate_enum_cases.md
```diff
@@ -0,0 +1,323 @@
...
+Note that as written, this would *not* allow C++ code to refer to `Foo::BAR`,
+`Foo::BAZ`, or `Foo::MULTI_WORD_ENUM`.
+
+#### Default `enum_case`
+
+Additionally, the same code would be emitted with either of the following:
+
+```
+enum Foo:
+  [$default (cpp) enum_case: kCamelCase"]
```
josh-conner, line 103: Nit: unbalanced quotes

doc/design_docs/alternate_enum_cases.md
```diff
@@ -0,0 +1,323 @@
...
+  BAR             = 1
+  BAZ             = 2
+  MULTI_WORD_ENUM = 4
+```
+
+or
+
+```
+[$default (cpp) enum_case: kCamelCase"]
```
josh-conner, line 112: Nit: unbalanced quotes

jasongraffius: Done

doc/design_docs/alternate_enum_cases.md
```diff
@@ -0,0 +1,323 @@
...
+enum Foo:
+  BAR             = 1  [(cpp) name: "kBar"]
+  BAZ             = 2  [(cpp) name: "kBaz"]
+  MULTI_WORD_ENUM = 4  [(cpp) name: "kMultiWordEnum"]
+```
+
+Note that the proposed `enum_case` design does not preclude an attribute of this
+nature for resolving other use-cases. Either multiple names could be emitted
+(those generated by `enum_case` as well as `name`), or under the principle of
+"specific overrides general" a `name`-like attribute could override any
```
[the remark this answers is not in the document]
jasongraffius: This is a great point, added the `name` attribute for future work and included this information.

doc/design_docs/alternate_enum_cases.md
jasongraffius: Ah, that's a great idea! Added to the future work section.

jasongraffius:
> I don't think of the constant syntax as an attribute of the data itself, but as an attribute of the environment where it will be used. Would it be possible to implement this as a command-line option?

I do agree that the constant syntax/style/case can vary based on the environment, but I'd say it's also part of the data itself for a few reasons:

 1. The existing `SHOUTY_CASE` already binds a name/style to the data implicitly.
 2. Alternatives like packed structs are inherently binding a name/style to the data, and would require modifying the source to alter. This is also true of other serialization languages like protobuf.
 4. There can be name collisions in some cases, like the `USB_1` vs `USB1` example, and it would be difficult to resolve this in a command line argument alone.

That being said, I think an approach to support something like being able to have a file in a "canonical" format/style but set/override the style at build time could still be supported in future work by allowing build-time command line options to set/add/override attributes in the AST or the IR after parsing. I can open a new issue for further discussion on that if we want to discuss going down that route.

reventlov:
> > I don't think of the constant syntax as an attribute of the data itself, but as an attribute of the environment where it will be used. Would it be possible to implement this as a command-line option?
> 
> I do agree that the constant syntax/style/case can vary based on the environment, but I'd say it's also part of the data itself for a few reasons:
> 
> 1. The existing `SHOUTY_CASE` already binds a name/style to the data implicitly.
> 2. Alternatives like packed structs are inherently binding a name/style to the data, and would require modifying the source to alter. This is also true of other serialization languages like protobuf.
> 3. There can be name collisions in some cases, like the `USB_1` vs `USB1` example, and it would be difficult to resolve this in a command line argument alone.
> 
> That being said, I think an approach to support something like being able to have a file in a "canonical" format/style but set/override the style at build time could still be supported in future work by allowing build-time command line options to set/add/override attributes in the AST or the IR after parsing. I can open a new issue for further discussion on that if we want to discuss going down that route.

I opened #85 for a feature that has been intended from when I first designed the attribute system: overlay files, so that you can add attributes without altering the original `.emb` file.  That is out of scope for this feature.

jasongraffius:
> I opened #85 for a feature that has been intended from when I first designed the attribute system: overlay files, so that you can add attributes without altering the original `.emb` file. That is out of scope for this feature.

Excellent, I agree the `.emboverlay` system would cover that use case. I'll leave this discussion out of this design doc in that case.

reventlov (approved):

PR #88, opened 2023-05-16
Fix header_generator_test

jasongraffius:
Due to a missing `unittest.main()` call, the header_generator_test was not actually running any tests when run with bazel/run directly. However, the failure was noticed by `python -m unittest discover`. Fixed the header_generator_test in the following ways:

 - Import test_util (tests failed due to missing import).
 - Change attribute_checker.normalize_and_verify to header_generator.generate_header.
 - Remove unused imports.
 - Make appropriate BUILD changes.
 - Add unittest.main() call so that unit tests are run.

Additionally, this change adds a (blank) \_\_init__.py to compiler/ and compiler/util/ so that they are importable by the unittest module. This means that running `python -m unittest discover -p '*_test.py'` from the main emboss repository should run all python tests in the compiler - and could be a good sanity check to make sure all python tests pass, even when `bazel test compiler/...` could miss tests (not declared in BUILD or missing the unittest.main() call).

### DIFF
--- a/compiler/front_end/BUILD
+++ b/compiler/front_end/BUILD
@@ -118,6 +118,7 @@ py_library(
     name = "test_util",
     testonly = 1,
     srcs = ["test_util.py"],
+    visibility = ["//:__subpackages__"],
     deps = [
     ],
 )
--- a/compiler/back_end/cpp/BUILD
+++ b/compiler/back_end/cpp/BUILD
@@ -55,6 +55,7 @@ py_test(
     deps = [
         ":header_generator",
         "//compiler/front_end:glue",
+        "//compiler/front_end:test_util",
     ],
 )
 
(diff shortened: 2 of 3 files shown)


compiler/front_end/BUILD
```diff
@@ -118,6 +118,7 @@ py_library(
     name = "test_util",
     testonly = 1,
     srcs = ["test_util.py"],
+    visibility = ["//:__subpackages__"],
```
jasongraffius, line 121: Note to reviewers: If it would be more appropriate to move this to `compiler/util` now that it is being used in both the front-end and the back-end (leaving only `glue` to be used directly from the front-end), I'd be happy to do that in this PR as well.

reventlov: Given what's in `test_util.py`, it does make sense to move it to `compiler/util`.

reventlov (commented):
Nice catch.

compiler/front_end/BUILD
jasongraffius: Sounds good! Moved to `compiler/util` in latest commit.

reventlov (approved):

PR #92, opened 2023-06-29
Emit compilation errors for bad C++ namespaces

jasongraffius:
Prior to this PR, an incorrectly-formatted namespace will fail an assert and print a python exception to the user. While this likely sufficient for a user to diagnose the issue, it is inconsistent with other error reporting in Emboss.

With this change, the namespace checking is implemented in the same way as other validations in the Emboss compiler, providing a specific file, line, and column range of the issue and printing the line containing the issue to the user.

### DIFF
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -66,6 +66,11 @@ _CPP_RESERVED_WORDS = set((
 # Emboss C++ support classes.
 _SUPPORT_NAMESPACE = "::emboss::support"
 
+# Regex matching a C++ namespace component. Captures component name.
+_NS_COMPONENT_RE = r"(?:^\s*|::)\s*([a-zA-Z_][a-zA-Z0-9_]*)\s*(?=\s*$|::)"
+# Regex matching a full C++ namespace (at least one namespace component).
+_NS_RE = fr"^\s*(?:{_NS_COMPONENT_RE})+\s*$"
+
 # TODO(bolms): This should be a command-line flag.
 _PRELUDE_INCLUDE_FILE = "runtime/cpp/emboss_prelude.h"
 
@@ -75,6 +80,22 @@ _SUPPORTED_ENUM_CASES = ("SHOUTY_CASE", "kCamelCase")
 # Verify that all supported enum cases have valid, implemented conversions.
 for _enum_case in _SUPPORTED_ENUM_CASES:
   assert name_conversion.is_case_conversion_supported("SHOUTY_CASE", _enum_case)
+
+
+def _get_namespace_components(namespace):
+  """Gets the components of a C++ namespace
+
+  Examples:
+    "::some::name::detail" -> ["some", "name", "detail"]
+    "product::name" -> ["product", "name"]
+    "simple" -> ["simple"]
+
+  Arguments:
+    namespace: A string containing the namespace. May be fully-qualified.
+
+  Returns:
+    A list of strings, one per namespace component."""
+  return re.findall(_NS_COMPONENT_RE, namespace)
 
 
 def _get_module_namespace(module):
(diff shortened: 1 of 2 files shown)


compiler/back_end/cpp/header_generator_test.py
```diff
@@ -266,6 +266,98 @@ def test_rejects_empty_enum_case(self):
...
+      ir = _make_ir_from_emb(test)
+      attr = ir.module[0].attribute[0]
+      self.assertEqual([[
+          error.error("m.emb", attr.value.source_location,
+                      'Bad namespace, does not match expected format.')
+      ]], header_generator.generate_header(ir)[1])
+
+  def test_rejects_empty_namespace(self):
+    for test in [
+        '[(cpp) namespace: ""]\n',
```
jasongraffius, line 322: Note: This test case (and only this completely-empty string test case) indicates a change in behavior, but the original behavior may have been unintended. That said, it seems that emboss' backwards-compatibility guarantees are fairly strict, so I can make this backwards compatible if that is required.

The previous behavior was to allow an empty namespace (but not one containing any spaces or just "::"), which would then be converted to the namespace `"emboss_generated_code"` (so it acts the same as if no namespace attribute had been specified at all).

compiler/back_end/cpp/header_generator_test.py
```diff
@@ -266,6 +266,98 @@ def test_rejects_empty_enum_case(self):
...
+        '[(cpp) namespace: "    "]\n',
+        '[(cpp) namespace: "  ::"]\n',
+        '[(cpp) namespace: ":: "]\n',
+        '[(cpp) namespace: " ::  "]\n',
+    ]:
+      ir = _make_ir_from_emb(test)
+      attr = ir.module[0].attribute[0]
+      self.assertEqual([[
+          error.error("m.emb", attr.value.source_location,
+                      'Bad namespace, does not match expected format.')
```
jasongraffius, line 334: It's pretty straightforward to add a "Empty namespace is not allowed" error message if that's preferred, could check against a regex of `"\s*(?:::)?\s*"` to use the empty message instead of the "bad namespace" message.

compiler/back_end/cpp/header_generator.py
```diff
@@ -1430,6 +1439,23 @@ def _offset_source_location_column(source_location, offset):
...
 
 
+def _verify_namespace_attribute(attr, source_file_name, errors):
+  if attr.name.text != attributes.Attribute.NAMESPACE:
+    return
+  namespace_value = attr.value.string_constant
+  if not re.match(_NS_RE, namespace_value.text):
+    errors.append([error.error(
+        source_file_name, namespace_value.source_location,
+        "Bad namespace, does not match expected format.")])
```
reventlov, line 1449: "Does not match expected format" is pretty vague; maybe some variation of `must be a valid C++ namespace, like "abc", "abc::def", or "::abc::def::ghi" (ISO/IEC 14882:2017 enclosing-namespace-specifier)`?

reventlov (changes requested):

compiler/back_end/cpp/header_generator_test.py
jasongraffius: That makes sense! I added this to the latest patch. If emboss ever needs to support emitting to the global namespace, it should be fairly simple to change this verification to accept that in the future.

compiler/back_end/cpp/header_generator_test.py
jasongraffius: Sounds good!

compiler/back_end/cpp/header_generator.py
jasongraffius: Done

reventlov (approved):
Thanks for this cleanup!

PR #112, opened 2024-02-27
Add --output-file embossc flag to allow setting exact generated file name

studgeek:
Allow users to specific the exact output header they want from embossc

### DIFF
--- a/embossc
+++ b/embossc
@@ -45,7 +45,12 @@ def _parse_args(argv):
   parser.add_argument("--output-path",
                       nargs=1,
                       default=".",
-                      help="Prefix to use for the generated output file.")
+                      help="""Prefix path to use for the generated output file.
+                              Defaults to '.'""")
+  parser.add_argument("--output-file",
+                      nargs=1,
+                      help="""File name to be used for the generated output
+                              file. Defaults to input_file suffixed by '.h'""")
   parser.add_argument("input_file",
                       type=str,
                       nargs=1,
@@ -54,6 +59,7 @@ def _parse_args(argv):
 
 
 def main(argv):
+
   flags = _parse_args(argv)
   base_path = os.path.dirname(__file__) or "."
   subprocess_environment = os.environ.copy()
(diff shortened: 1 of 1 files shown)


embossc
```diff
@@ -46,6 +46,10 @@ def _parse_args(argv):
                       nargs=1,
                       default=".",
                       help="Prefix to use for the generated output file.")
+  parser.add_argument("--output-file",
+                      nargs=1,
+                      help="""Exact file path to use for the generated output
+                              file. Overrides --output-path is ignored.""")
```
reventlov, line 52: > Overrides --output-path is ignored.

"Only one of --output-path and --output-file may be used", maybe?  And update the --output-path message and throw an error if both are specified.

reventlov (changes requested):

reventlov:
`--output-file` is consistent with `--output-path` and the other `embossc` flags.

embossc
AaronWebster: That works for me.  And you are correct, there is inconsistent dash/underscore used here which is beyond the scope of this PR, so disregard that comment.

studgeek: I reworked closer to what Aaron suggested since it seems more straightforward relative to existing behavior and  flags. Now  --output-file help is "File name to be used for the generated output file. Defaults to input_file suffixed by '.h'". PTAL

embossc
```diff
@@ -96,10 +102,15 @@ def main(argv):
...
 
-  output_file = os.path.join(flags.output_path[0], flags.input_file[0]) + ".h"
-  os.makedirs(os.path.dirname(output_file), exist_ok=True)
+  if flags.output_file:
+    output_file = flags.output_file[0]
+  else:
+    output_file = flags.input_file[0] + ".h"
+
+  output_filepath = os.path.join(flags.output_path[0], output_file)
+  os.makedirs(os.path.basename(os.path.dirname(output_filepath)), exist_ok=True)
```
reventlov, line 111: Why did you add the `basename` call?  That's a breaking behavioral change.

reventlov (changes requested):

embossc
studgeek: Ah, that is a merge mistake on my part. Removed.

reventlov (approved):

Issue #119, opened 2024-03-29
Improve performance of `ir_traverse._call_with_optional_args`


EricRahm:
### Overview
As a follow-up to the analysis of #118 I found that the `inspect.getfullargspec` in `_call_with_optional_args` is taking about 10% of runtime:
![image](https://github.com/google/emboss/assets/426919/21ebdc0b-fb73-414b-92aa-1dacf5247c14)

A proof-of-concept fully removed this overhead.

### Background
This call is invoked as part of the `fast_traverse_ir_top_down` implementation. The `action` and `incidental_action` operations are all run through this helper function which is invoked about 100,000 times in my test. Currently `inspect.getfullargspec` is called every time. Additionally the argument extraction does a fair amount of list creation and iteration that could also be avoided if memoized:

```python
def _call_with_optional_args(function, positional_arg, keyword_args):
  """Calls function with whatever keyword_args it will accept."""
  argspec = inspect.getfullargspec(function)
  if argspec.varkw:
    # If the function accepts a kwargs parameter, then it will accept all
    # arguments.
    # Note: this isn't technically true if one of the keyword arguments has the
    # same name as one of the positional arguments.
    return function(positional_arg, **keyword_args)
  else:
    ok_arguments = {}
    for name in keyword_args:
      if name in argspec.args[1:] or name in argspec.kwonlyargs:
        ok_arguments[name] = keyword_args[name]
    for name in argspec.args[1:len(argspec.args) - len(argspec.defaults or [])]:
      assert name in ok_arguments, (
          "Attempting to call '{}'; missing '{}' (have '{!r}')".format(
              function.__name__, name, list(keyword_args.keys())))
    return function(positional_arg, **ok_arguments)
```

### Possible Fix
Memoize the extracted lists of valid and required args from `inspect.getfullargspec`for a given function. Something like:
```python
@simple_memoizer.memoize
def get_function_args(function):
  argspec = inspect.getfullargspec(function)
  return {
     "valid": argspec.args[1:] or name in argspec.kwonlyargs
     "required": argspec.args[1:len(argspec.args) - len(argspec.defaults or [])]
  }
```

### Drawbacks
We currently use some `lambda` functions in addition to static methods for actions. We'd need to convert those to static methods, otherwise we'd end up with a memoization entry for each dynamic lambda instance.

EricRahm:
#120 fixed this.

PR #120, opened 2024-03-29
Improve perf of `_call_with_optional_args`

EricRahm:
Fixes issue #119.

1. In preparation for a perf improvement all actions that use lambdas are converted to static functions. Additionally duplicate actions are removed.
2. The results of `getfullargspec` are memoized and encapsulated in a helper class.

This gets us from 100K calls to `getfullargspec` to about 60 and reduces overhead from 10% to ~0% in my tests.

### DIFF
--- a/compiler/util/traverse_ir.py
+++ b/compiler/util/traverse_ir.py
@@ -17,27 +17,98 @@
 import inspect
 
 from compiler.util import ir_pb2
+from compiler.util import simple_memoizer
+
+
+class _FunctionCaller:
+  """Provides a template for setting up a generic call to a function.
+
+  The function parameters are inspected at run-time to build up a set of valid
+  and required arguments. When invoking the function unneccessary parameters
+  will be trimmed out. If arguments are missing an assertion will be triggered.
+
+  This is currently limited to functions that have at least one positional
+  parameter.
+
+  Example usage:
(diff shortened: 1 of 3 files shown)


compiler/util/traverse_ir.py
```diff
@@ -17,27 +17,98 @@
 import inspect
 
 from compiler.util import ir_pb2
+from compiler.util import simple_memoizer
+
+
+class _FunctionCaller:
```
reventlov, line 23: It would be a more invasive change, but you could just require that all action functions have a `**kwargs` parameter, and skip all the reflection.

reventlov (approved):

compiler/util/traverse_ir.py
EricRahm: That's a good point, I did consider it but as you noted:
1. It's more invasive
2. I found the error messages if you forget to do that a bit harder to understand

Going from 100K calls to 60 or so seems a reasonable tradeoff for perf and dev productivity.

reventlov: Only 60 calls to `getfullargspec`, but still a nontrivial amount of work on each invocation to filter the argument lists.

Either way: it's a big improvement over the status quo.

PR #123, opened 2024-04-03
Call parsing and generation functions directly in `embossc`

### LINKED ISSUE #122: `embossc` unnecessarily spawns 2 subprocesses
As part of a series of end-to-end performance investigations (see #118, #119) I found that the `embossc` script is [spawning a subprocess for the front-end step and then the back-end steps](https://github.com/google/emboss/blob/a5126d648a1d32c4049ef3f5e19122a74cc78c5e/embossc#L84-L100). This adds a small overhead of launching the subprocesses, but moreover requires serializing, piping between processes, and deserializing IR to an intermediate JSON representation.

For a larger emb a PoC showed a reduction of 16% overhead by removing the subprocesses. I'm proposing going from:
```python
  front_end_status = subprocess.run(front_end_args,
                                    stdout=subprocess.PIPE,
                                    env=subprocess_environment)

  if front_end_status.returncode != 0:
    return front_end_status.returncode

  back_end_status = subprocess.run(
    [
      sys.executable,
      os.path.join(base_path, "compiler", "back_end", "cpp",
        "emboss_codegen_cpp.py"),
    ],
    input=front_end_status.stdout,
    stdout=subprocess.PIPE,
    env=subprocess_environment
  )
```
to something like:
```python
    ir, _, errors = glue.parse_emboss_file(
        flags.input_file[0], emboss_front_end._find_in_dirs_and_read(flags.import_dirs))
    if not errors:
      header, errors = header_generator.generate_header(ir)
```

EricRahm:
Rather than spawning 2 subprocesses and serializing IR data between them
we can just use the front end and back end libraries directly. This is a
code simplification as well as a performance improvement showing up a 3%
reduction in overhead for larger emb files.

Fixes #122.

### DIFF
--- a/embossc
+++ b/embossc
@@ -18,9 +18,7 @@
 
 import argparse
 import os
-import subprocess
 import sys
-
 
 def _parse_args(argv):
   parser = argparse.ArgumentParser(description="Emboss compiler")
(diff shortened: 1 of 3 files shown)


embossc
```diff
@@ -59,48 +57,28 @@ def _parse_args(argv):
 
 
 def main(argv):
-
   flags = _parse_args(argv)
   base_path = os.path.dirname(__file__) or "."
-  subprocess_environment = os.environ.copy()
+  sys.path.append(base_path)
```
robrussell, line 62: Can the imports be moved to the top? Afaict these modules are imported elsewhere without updating the path.

robrussell (changes requested):

embossc
EricRahm: Unfortunatly not. `embossc` is the entry point for end-users of the `emboss`, for example:
```bash
python3 ~/workspace/emboss/embossc
  --output-path /tmp \
  --import-dir public \
  public/pw_bluetooth/hci_events.emb
```
The above will fail if we move the imports to before we set up the python path. 

We can see the previous did essentially the same thing, just with `PYTHONPATH` instead:
```
  if subprocess_environment.get("PYTHONPATH"):
    subprocess_environment["PYTHONPATH"] = (
      base_path + ":" + subprocess_environment.get("PYTHONPATH"))
  else:
    subprocess_environment["PYTHONPATH"] = base_path
```

Issue #116, opened 2024-03-28
Emboss' use of `IsChar<>` is incorrect


reventlov:
Internally, Emboss uses a template type called `IsChar<>` to determine whether a type is *safe to use with aliasing operations*.

This is based on a misreading of the standard: only `char` and `unsigned char` (and, as of C++17, `std::byte`) can safely alias other types, but `IsChar<signed char>::value` is `true`.  The type `signed char` is not alias-safe.

`IsChar` is also used to determine which types are acceptable to pass into view constructors, so removing `signed char` from that list is technically a breaking change; however, `signed char` (and `std::int8_t`, which is an alias of `signed char` on any real-world compiler that uses an unsigned plain `char` type) is rarely used in this context, so it seems unlikely that this would break any actual uses.

The new (... 7 years ago) type `std::byte` should also be added to the list, though it needs to be guarded by `#if __cplusplus >= 201703L` until at least 2027.

And given that it should not be `true` for all character types and only character types, `IsChar` should probably be renamed to `IsAliasSafe` or similar.

PR #127, opened 2024-04-05
Stop using std::basic_string<unsigned char>

tpudlik:
std::basic_string<unsigned char> relies on a non-standard generic char_traits<> implementation, recently removed from libc++.

Upstreams cl/605337651.

Relevant to #116.


tpudlik:
@robrussell

robrussell (approved):

Issue #125, opened 2024-04-03
Improve perf of C++ backend


EricRahm:
As part of my ongoing analysis of emboss performance (see #118, #119 #122) I took a look at profiling just the C++ backend. The same methodology applies.

### Analysis
![image](https://github.com/google/emboss/assets/426919/7a45117d-6f86-45ba-8e80-9b25a4fb0c1d)

- We can see that 75% of the time is spent deserializing IR data. #122 should cover fixing this when running `embossc`, but it will still be an issue if the backend is run standalone. #118 will help in that case.
- 24% of the time is spent generating headers

Drilling in a little more we can see that 13% of time is spent formatting the code templates:
![image](https://github.com/google/emboss/assets/426919/c5f3f56e-da49-40d7-a05c-0717f7a90d2a)

### Proposal
1. Ignore the deserialization overhead for now, other issues should cover that
2. Replace our [bespoke template formatting](https://github.com/google/emboss/blob/a5126d648a1d32c4049ef3f5e19122a74cc78c5e/compiler/back_end/util/code_template.py) with `string.Template`. A proof-of-concept shows that reduces overhead by *11%*.
![image](https://github.com/google/emboss/assets/426919/77915b73-2228-4117-b1db-ad6ee026d396)

PR #130, opened 2024-04-10
Switch `code_templates` to `string.Template`

EricRahm:
Python's standard `string.Template` is significantly more performant than the custom `code_template` formatter. At the cost of removing custom formatting support (unused) and changing our format string to `${foo}` instead of `$_foo_$` we can achieve an 11% speedup in the back end code generation for C++.

Fixes: #125
Test: bazel test 'compiler/back_end/cpp:*'


EricRahm:
FYI @robrussell

compiler/back_end/util/code_template.py
```diff
@@ -22,59 +22,7 @@
...
-      after_last_delimiter = match.end()
-    yield format_string[after_last_delimiter:], None, None, None
-
-
-_FORMATTER = _CppFormatter()
-
-
-def format_template(template, *args, **kwargs):
+def format_template(template, **kwargs):
   """format_template acts like str.format, but uses $_name_$ instead of {name}.
```
robrussell, line 26: Doc string still describes old function.

compiler/back_end/util/code_template_test.py
```diff
@@ -14,85 +14,85 @@
...
   """Tests for code_template.format."""
 
   def test_no_replacement_fields(self):
-    self.assertEqual("foo", code_template.format_template("foo"))
-    self.assertEqual("{foo}", code_template.format_template("{foo}"))
-    self.assertEqual("$foo$", code_template.format_template("$foo$"))
-    self.assertEqual("$_foo$", code_template.format_template("$_foo$"))
-    self.assertEqual("$foo_$", code_template.format_template("$foo_$"))
+    self.assertEqual("foo", _format_template_str("foo"))
+    self.assertEqual("{foo}", _format_template_str("{foo}"))
```
robrussell, line 29: If the string `${foo}` is needed then I think that's

```
self.assertEqual("{foo}", _format_template_str("{foo}"))
```

Based on https://docs.python.org/3/library/string.html#template-strings . The extra test would be useful to inform readers who need that escaping.

compiler/back_end/util/code_template_test.py
```diff
@@ -14,85 +14,85 @@
...
-    self.assertEqual("'foo'",
-                     code_template.format_template("$_bar!r_$", bar="foo"))
-    self.assertEqual("==foo==",
-                     code_template.format_template("$_bar:=^7_$", bar="foo"))
-    self.assertEqual("=='foo'==",
-                     code_template.format_template("$_bar!r:=^9_$", bar="foo"))
-    self.assertEqual("xx=='foo'==yy",
-                     code_template.format_template("xx$_bar!r:=^9_$yy",
-                                                   bar="foo"))
+    # Basic string.Templates don't support formatting values.
```
robrussell, line 41: There have been a few different ways to do formatting in Python so maybe I'm mistaken but I thought the tests from line 43-51 should still work.

compiler/back_end/util/code_template.py
```diff
@@ -22,59 +22,7 @@
 import string
 
 
-class _CppFormatter(string.Formatter):
-  """Customized Formatter using $_name_$ instead of {name}.
```
robrussell, line 26: I'm unclear on why `$_name_$` is used instead of `{name}` but it does mean that the templates might be syntactically valid C++. This makes it easier to edit the templates with syntax highlighting in an IDE. Maybe there's some way to get that back in the future but nothing comes to mind.

robrussell (commented):
There are additional references to `$_name_$` syntax in comments in [compiler/back_end/cpp/header_generator.py]( https://github.com/google/emboss/blob/0a362eabb90eb57b5036955f0ffbdc1d9594bf36/compiler/back_end/cpp/header_generator.py#L984) that should be revised.

*Probably* not relevant but the sequence `${name}` already shows up in [emboss_defines.h](https://github.com/google/emboss/blob/0a362eabb90eb57b5036955f0ffbdc1d9594bf36/runtime/cpp/emboss_defines.h#L31).

compiler/back_end/util/code_template.py
EricRahm: Updated.

compiler/back_end/util/code_template_test.py
EricRahm: This maybe didn't render correctly but I think you want me to test `$${foo}`, if so added that.

compiler/back_end/util/code_template_test.py
EricRahm: They do not (the following line shows that a `ValueError` is raised to make that clear).

compiler/back_end/util/code_template.py
EricRahm: I've been editing with VSCode with these changes, syntax highlighting is okay. Maybe this is an issue for specific IDEs.

image

EricRahm:
Rebased and updated per the code review comments, thanks for taking a look!

robrussell (approved):

Issue #118, opened 2024-03-28
Improve performance of `ir_pb2`


EricRahm:
### Pitch
Modernize and simplify the `ir_pb2` data structures to improve performance and maintainability.

I propose working from two angles:
1. Reduce the total number of IR data class mutations
2. Make mutations quicker

I have a proof-of-concept in place that reduces runtime by about ~30%.

### Approaches
1. Perform some basic cleanup and modernization
    1. Remove legacy protobuf naming: Move ir_pb2 -> ir_data
    2. Pull out enumeration-like fields to proper python `enums.Enum`
2. Simplify the IR data classes to just hold data:
    1. Remove `Optional`, `Repeated`, etc in favor of typed fields `dataclasses`
    2. Add helpers to integrate dataclass `Field`s with existing psuedo-field-spec support
4. Move dynamic building and serialization to separate classes and functions
    1. Add an explicit `builder` to opt-in to mutation
    4. Add an explicit `reader` to allow chained-reads of potentially unset fields 

An example of what I'm thinking for a converted `Function` class:
```python
class FunctionMapping(enums.Enum):
  UNKNOWN = 0
  ADDITION = 1           # +
  SUBTRACTION = 2        # -
  MULTIPLICATION = 3     # *
  EQUALITY = 4           # ==
  INEQUALITY = 5         # !=
  AND = 6                # &&
  OR = 7                 # ||
  LESS = 8               # <
  LESS_OR_EQUAL = 9      # <=
  GREATER = 10           # >
  GREATER_OR_EQUAL = 11  # >=
  CHOICE = 12            # ?:
  MAXIMUM = 13           # $max()
  PRESENCE = 14          # $present()
  UPPER_BOUND = 15       # $upper_bound()
  LOWER_BOUND = 16       # $lower_bound()

@dataclasses.dataclass
class Function:
  function: typing.Optional[FunctionMapping]
  args: list[Expression]
  function_name: typing.Optional[Word]
  source_location: typing.Optional[Location]
```
vs the current style
```python
@message
class Function(Message):
  """IR for a single function (+, -, *, ==, $max, etc.) in an expression."""
  UNKNOWN = 0
  ADDITION = 1           # +
  SUBTRACTION = 2        # -
  MULTIPLICATION = 3     # *
  EQUALITY = 4           # ==
  INEQUALITY = 5         # !=
  AND = 6                # &&
  OR = 7                 # ||
  LESS = 8               # <
  LESS_OR_EQUAL = 9      # <=
[...]

PR #131, opened 2024-04-12
Split out IR enums

EricRahm:
Adds `FunctionMapping` and `AddressableUnit` enums and pulls out the existing definitions from the IR classes. This is in preparation for #118.

### DIFF
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -250,8 +250,8 @@ def _get_adapted_cpp_buffer_type_for_field(type_definition, size_in_bits,
                                            buffer_type, byte_order,
                                            parent_addressable_unit):
   """Returns the adapted C++ type information needed to construct a view."""
-  if (parent_addressable_unit == ir_pb2.TypeDefinition.BYTE and
-      type_definition.addressable_unit == ir_pb2.TypeDefinition.BIT):
+  if (parent_addressable_unit == ir_pb2.AddressableUnit.BYTE and
+      type_definition.addressable_unit == ir_pb2.AddressableUnit.BIT):
     assert byte_order
     return _bytes_to_bits_convertor(buffer_type, byte_order, size_in_bits)
   else:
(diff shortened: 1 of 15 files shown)


compiler/back_end/cpp/header_generator.py
```diff
@@ -374,7 +374,7 @@ def _get_cpp_view_type_for_physical_type(
             element_view_parameter_types="".join(
                 ", " + p for p in element_view_parameter_types),
             element_size=element_size,
-            addressable_unit_size=parent_addressable_unit,
+            addressable_unit_size="1" if parent_addressable_unit == ir_pb2.AddressableUnit.BIT else "8",
```
robrussell, line 377: `ir_pb2.AddressableUnit` inherits from int, are the "1" and "8" needed or can this be `str(addressable_unit_size)` or just `addressable_unit_size` directly?

robrussell (approved):

compiler/back_end/cpp/header_generator.py
EricRahm: Ah okay, so the deal is without this we get the literal string `AddressableUnit.BYTE` in our generated headers instead of `8`. This isn't related to #130. I did find a small backwards compat issue that I'm going to upload where legacy serialization supports the string `BYTE` in json. This isn't an issue later on in my patch series but I'd like the tests to not fail!

compiler/back_end/cpp/header_generator.py
```diff
@@ -374,7 +374,7 @@ def _get_cpp_view_type_for_physical_type(
             element_view_parameter_types="".join(
                 ", " + p for p in element_view_parameter_types),
             element_size=element_size,
-            addressable_unit_size=parent_addressable_unit,
+            addressable_unit_size=int(parent_addressable_unit),
```
EricRahm, line 377: Switched this to explicitly convert to `int` to be clear about what we're doing.

compiler/util/ir_pb2.py
```diff
@@ -107,9 +108,12 @@ def set(self, obj, value):
       self._set_value(obj, self._type(**value))
     elif isinstance(value, _Text) and self._decode_names:
       self._set_value(obj, self._type(self._decode_names(value)))
+    elif isinstance(value, _Text) and issubclass(self._type, enum.Enum):
+      self._set_value(obj, getattr(self._type, value))
```
EricRahm, line 112: Added this path to handle the case where a json serialization had a string representation, eg `BYTE` instead of `8`

EricRahm:
@robrussell Should be good to go now.

Issue #128, opened 2024-04-08
C++ backend generates excessively large headers


EricRahm:
### Issue
For a ~75KB `emb` file I'm seeing a ~7.5MB header generated. An example for the `pw_bluetooth` project:

|Definition File|Size (KB)|Header Size (KB)|
|-------------|-----|------------|
| hci_events.emb | 76 | 7488 |
| hci_commands.emb | 108 | 7376 |
| hci_vendor.emb | 32 | 2328 |
| hci_common.emb | 32 | 824 |
| hci_data.emb | 4 | 204 |
| l2cap_frames.emb | 4 | 60 |
| hci_h4.emb | 4 | 8 |

The scaling of input size to output size isn't consistent, so the behavior is definitely content driven.

### Impact
- Overhead of the backend generation step: larger headers take longer to generate
- Slower compilation times: large headers result in long compile times even if a majority of the header is unreferenced
- Hard to navigate: consumers of the headers have a lot of code to look at
- Code analysis tools like `clangd` slow to a crawl when encountering the generated headers

### Possible Improvements
1. C++ fragment templates include comments relevant to the emboss C++ backend, but not the generated code itself. We could trim out the comments when generating the code
2. Large amount of text conversion code (ostream, enum strings, etc). For many projects this is unnecessary, we should make it optional and/or split out to a separate header.
3. Large amount of code generated to wrap / validate / extract constant values, we can just use the constants directly.

EricRahm:
I have a few patches pending that'll address 1 and 2. Just need to hold off until #130 lands since it has a significant changes to the code templates. FYI @robrussell

PR #133, opened 2024-04-18
Strip comments from templates

EricRahm:
This removes comments related to the emboss backend code generator from
the generated emboss headers. For an example project we see several MB
of reduction in size of the generated headers:

|File              | src (KB) | .h (KB) before | .h (KB) after |
|------------------|----------|----------------|---------------|
| hci_events.emb   |       76 |           7488 |          5356 |
| hci_commands.emb |      108 |           7376 |          5612 |
| hci_vendor.emb   |       32 |           2328 |          1748 |
| hci_common.emb   |       32 |            824 |           644 |
| hci_data.emb     |        4 |            204 |           152 |
| l2cap_frames.emb |        4 |             60 |            48 |
| hci_h4.emb       |        4 |              8 |             8 |

The comments are stripped while generating the code fragment templates,
this allows us to keep comments useful to emboss development in-place
while reducing the impact on generated code size.

This is part of resolving #128


EricRahm:
Note this is stacked on #130, only the patch titled `Strip comments from temlates` is relevant.

EricRahm:
FYI @robrussell.

compiler/back_end/util/code_template.py
```diff
@@ -66,6 +66,7 @@ def parse_templates(text):
     A namedtuple object whose attributes are the templates from text.
```
robrussell, line 66: Docs should mention that full line comments starting with `//` are removed. It won't matter in most cases but since this uses regex there are weird things like
```
    /*
    some_code();
    // */
```
that would behave differently after stripping.

robrussell (approved):

compiler/back_end/util/code_template.py
studgeek: I didn't know you could do that :) (have always assumed line would take precedence for whole line). Perhaps we could update the regex to explicitly not remove that case - line comments that contain a closing block comment?

In addition to updating docs?

robrussell (approved):

PR #137, opened 2024-04-18
Bump required version to 3.9 and add it to bazel

### LINKED ISSUE #136: emboss fails to run on python 3.8
Running emboss is currently broken in python version 3.8. I believe our code now relies on at least 3.9 as of PR #111. We could try to add back 3.8 support but given it's EOL in a few months and #111 seems to be fixing a real issue I'd suggest just bumping the min version to 3.9.

STR:

1. Verify the a new default python3 works
```
# run with local installation
$ python3 --version
Python 3.11.8
$ bazel test \
    'compiler/back_end/cpp:*' \
    'compiler/front_end:*' 'compiler/util:*' 'integration/googletest:*'

<snip>

Executed 80 out of 80 tests: 80 tests pass.
//integration/googletest:emboss_test_util_test                           PASSED in 0.2s

Executed 80 out of 80 tests: 80 tests pass.
```
2. Explicitly run against python 3.8
```
$ pyenv install 3.8
$ bazel test --action_env=PYENV_VERSION=3.8 \
    'compiler/back_end/cpp:*' 'compiler/front_end:*' \
    'compiler/util:*' 'integration/googletest:*'

<snip>

//compiler/front_end:write_inference_test                                FAILED in 2.7s
  /usr/local/google/home/erahm/.cache/bazel/_bazel_erahm/ce8d987ff54be8a48c8957a8dbac668b/execroot/com_google_emboss/bazel-out/k8-fastbuild/testlogs/compiler/front_end/write_inference_test/test.log

Executed 80 out of 80 tests: 66 tests pass and 14 fail locally.

$ cat /usr/local/google/home/erahm/.cache/bazel/_bazel_erahm/ce8d987ff54be8a48c8957a8dbac668b/execroot/com_google_emboss/bazel-out/k8-fastbuild/testlogs/compiler/front_end/write_inference_test/test.log
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //compiler/front_end:write_inference_test
-----------------------------------------------------------------------------
EEEEEEEEEEEEEE
======================================================================
ERROR: test_adds_alias_write_method_to_alias_of_alias_of_physical_field (__main__.WriteInferenceTest)
----------------------------------------------------------------------

<snip>

  File "/tmp/bazel-working-directory/com_google_emboss/bazel-out/k8-fastbuild/bin/compiler/front_end/write_inference_test.runfiles/com_google_emboss/compiler/util/resources.py", line 21, in load
    with importlib.resources.files(
AttributeError: module 'importlib.resources' has no attribute 'files'
```
3. Explicitly run against 3.9, note things work again
```
$ pyenv install 3.9
$ bazel test --action_env=PYENV_VERSION=3.9 \
    'compiler/back_end/cpp:*' 'compiler/front_end:*' \
    'compiler/util:*' 'integration/googletest:*'
<snip>
//integration/googletest:emboss_test_util_test                           PASSED in 0.0s

Executed 80 out of 80 tests: 80 tests pass.
```

EricRahm:
This bumps the required version to 3.9 which is what our codebase currently requires. It also adds a bazel rule to install a hermetic python 3.9 package. This will make `bazel` always use the hermetic package which should avoid backwards compatibility issues in the future.

Fixes #136

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -31,6 +31,14 @@ http_archive(
     url = "https://github.com/bazelbuild/rules_python/releases/download/0.31.0/rules_python-0.31.0.tar.gz",
 )
 
-load("@rules_python//python:repositories.bzl", "py_repositories")
+load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
 
 py_repositories()
+
+# Use Python 3.9 for bazel Python rules.
+python_register_toolchains(
+    name = "python3",
+    python_version = "3.9",
+)
+
+load("@python3//:defs.bzl", "interpreter")
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -24,7 +24,7 @@ emails.
 
 #### Running the Emboss Compiler
 
-The Emboss compiler requires Python 3.8 or later -- the minimum supported
+The Emboss compiler requires Python 3.9 or later -- the minimum supported
 version tracks the support timeline of the Python project.  On a Linux-like
 system with Python 3 installed in the usual place (`/usr/bin/python3`), you
 can run the embossc script at the top level on an `.emb` file to generate


EricRahm:
FYI @AaronWebster @robrussell

WORKSPACE
```diff
@@ -31,6 +31,14 @@ http_archive(
...
 
-load("@rules_python//python:repositories.bzl", "py_repositories")
+load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
 
 py_repositories()
+
+# Use Python 3.9 for bazel Python rules.
+python_register_toolchains(
+    name = "python3",
+    python_version = "3.9",
```
robrussell, line 41: Without this change which version of Python is used? Would it default to current (i.e. 3.11 today)?

EricRahm: It's arbitrary, it'll use whatever the user has on their system. I happen to have `3.11.8` on my dev machine.

robrussell: Is there a way for users to get that functionality again after this change? Can it be overridden with a flag to bazel, for example? I'd assume that 3.11 would generally have better performance so a user might rather keep using that. If there's a flag or some way to override then a comment would be sufficient.

EricRahm: Hmmm good question, while testing things out I used the `--action_env=PYENV_VERSION=3.8` param to bazel so presumably something like that might work. For my current use case we're using a cusom integration with `gn` and `cmake` so they'll just use whatever we've specified in our build system. For bazel I'm not sure which toolchain would "win.", I'll do some research here, but I suspect the toolchain we specify in our project would be used.

Really my intention was to just use this for tests and not the exported emboss rules, I'll see if I can scope this to just that use case.

EricRahm:
@reventlov fyi if you have an opinion

reventlov:
> @reventlov fyi if you have an opinion

No strong opinion.  Scoping to tests only would use fewer system resources while compiling, but leave the open possibility of things breaking if an end user has Python 3.8 (or earlier) as their system Python 3.

WORKSPACE
jasongraffius: If we want to keep the current behavior (use the system python), but still ensure that things aren't broken on the oldest supported Python release, I could add `setup-python` to the GitHub actions, add a `.python-version` file to the repo. That would be the "source of truth" so to speak of the minimum supported version for the repo and we would just bump that whenever we update the supported version of python.

jasongraffius:
As I just discovered in #167, this is insufficient to fix the issue as currently 3.9 is also failing. I confirmed on all other supported Python versions and it works fine, but it is failing on both 3.8 and 3.9.

EricRahm:
This should "fix" the 3.8 issues and highlight failures on 3.9 so that we prevent future regressions right? Ofc it's already regressed, but once that's fixed it should do the job right?

jasongraffius:
Yeah, I didn't word that well, I meant without fixing 3.9 first.

jasongraffius:
Would it be possible to rebase this on #163 and do this or something similar to this in the `MODULE.bazel` file instead of `WORKSPACE`? That way we can land #163 and this without conflict.

EricRahm:
Im fine landing #163 and breaking this PR. I'll be out a few weeks so if
you want to land a reworked version that's fine with me and we can just
abandon this one.

EricRahm:
Lets retire this, I think we're at the point where we should  bump to 3.10.

PR #141, opened 2024-05-03
Rename ir_pb2 to ir_data

EricRahm:
The IR classes are no longer based on a protobuf implementation. This makes that clear by migrating to a more neutral `ir_data` name. This will also help avoid tripping up static analysis that looks for protobuf-like things.

Part of #118.


EricRahm:
FYI @robrussell

PR #142, opened 2024-05-03
Split out an `IrDataSerializer` class

EricRahm:
In preparation for reworking `ir_data` move json serialazition to it's own class. For now this is simply a wrapper. All `to_json` call sites are updated.

Part of #118.


EricRahm:
fyi @AaronWebster @robrussell

robrussell (approved):

PR #143, opened 2024-05-14
Add `ir_data_utils.builder`

EricRahm:
This adds an `ir_data_utils.builder` wrapper that will be used to dynamically build up `ir_data` instances rather than the ad-hoc behavior of `ir_data` field access. For now this is a no-op and simply annotates sites that are mutating `ir_data` instances.

Part of #118.

### DIFF
--- a/compiler/front_end/module_ir.py
+++ b/compiler/front_end/module_ir.py
@@ -26,6 +26,7 @@ import re
 import sys
 
 from compiler.util import ir_data
+from compiler.util import ir_data_utils
 from compiler.util import name_conversion
 from compiler.util import parser_types
 
@@ -793,9 +794,9 @@ def _type_definition(type_definition):
 def _structure(struct, name, parameters, colon, comment, newline, struct_body):
   """Composes the top-level IR for an Emboss structure."""
   del colon, comment, newline  # Unused.
-  struct_body.structure.source_location.start.CopyFrom(
+  ir_data_utils.builder(struct_body.structure).source_location.start.CopyFrom(
       struct.source_location.start)
-  struct_body.structure.source_location.end.CopyFrom(
+  ir_data_utils.builder(struct_body.structure).source_location.end.CopyFrom(
       struct_body.source_location.end)
   struct_body.name.CopyFrom(name)
   if parameters.list:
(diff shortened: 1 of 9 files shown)


EricRahm:
Fyi @robrussell @AaronWebster

compiler/front_end/module_ir.py
```diff
@@ -793,9 +794,9 @@ def _type_definition(type_definition):
 def _structure(struct, name, parameters, colon, comment, newline, struct_body):
   """Composes the top-level IR for an Emboss structure."""
   del colon, comment, newline  # Unused.
-  struct_body.structure.source_location.start.CopyFrom(
+  ir_data_utils.builder(struct_body.structure).source_location.start.CopyFrom(
```
robrussell, line 797: With the current implementation of `builder` this has the same effect as before the change. But the comment on the `builder` implementation implies that this will change. Normally I'd expect something along the lines of

```
builder = ir_data_utils.builder(struct_body.structure) ...
builder.source_location.end ...
return builder.build()
```

I guess it seems odd that the return value appears to be ignored. Or will the future `builder()` impl still return something equivalent to `self`?

compiler/util/ir_data_utils.py
```diff
@@ -29,3 +29,11 @@ def to_json(self, *args, **kwargs):
   def from_json(data_cls: type[ir_data.Message], data):
     """Constructs an IR data class from the given JSON string"""
     return data_cls.from_json(data)
+
+
+def builder(target: ir_data.Message) -> ir_data.Message:
+  """Provides a wrapper for building up IR data classes.
+
+  This is a no-op and just used for annotation for now.
```
robrussell, line 37: Can the `for now` be more specific? Maybe just add a link to #118 .

EricRahm: I have all the patches ready, ideally we'd just land them somewhat quickly and not bother but if it's going to take an extended amount of time we can update it.

compiler/front_end/module_ir.py
EricRahm: We could maybe choose a different name, it's not exactly a builder in the "builder pattern" sense. The intention of the `builder` is to let you create a field by accessing it (the next patch will be adding a `reader` that lets you access a non-existent field, but it won't actually create it).

PR #145, opened 2024-05-31
Add `ir_data_utils.copy` and `update`

EricRahm:
This adds `ir_data_utils` helpers to handle copying and updating `ir_data` instances. Some instances of `CopyFrom` that aren't associated with a `builder` are updated to use `copy` or `update`.

Part of #118.

### DIFF
--- a/compiler/front_end/module_ir.py
+++ b/compiler/front_end/module_ir.py
@@ -150,7 +150,10 @@ def _really_build_ir(parse_tree, used_productions):
     used_productions.add(parse_tree.production)
     result = _handlers[parse_tree.production](*parsed_children)
     if parse_tree.source_location is not None:
-      result.source_location.CopyFrom(parse_tree.source_location)
+      if result.source_location:
+        ir_data_utils.update(result.source_location, parse_tree.source_location)
+      else:
+        result.source_location = ir_data_utils.copy(parse_tree.source_location)
     return result
   else:
     # For leaf nodes, the temporary "IR" is just the token.  Higher-level rules
@@ -798,7 +801,10 @@ def _structure(struct, name, parameters, colon, comment, newline, struct_body):
       struct.source_location.start)
   ir_data_utils.builder(struct_body.structure).source_location.end.CopyFrom(
       struct_body.source_location.end)
-  struct_body.name.CopyFrom(name)
+  if struct_body.name:
+    ir_data_utils.update(struct_body.name, name)
+  else:
+    struct_body.name = ir_data_utils.copy(name)
   if parameters.list:
     struct_body.runtime_parameter.extend(parameters.list[0].list)
   return struct_body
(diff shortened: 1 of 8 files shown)


EricRahm:
FYI @robrussell

compiler/front_end/module_ir.py
```diff
@@ -1166,7 +1171,10 @@ def _enum_value_body(indent, docs, attributes, dedent):
 def _external(external, name, colon, comment, newline, external_body):
   del colon, comment, newline  # Unused.
   ir_data_utils.builder(external_body.source_location).start.CopyFrom(external.source_location.start)
-  external_body.name.CopyFrom(name)
+  if external_body.name:
+    ir_data_utils.update(external_body.name, name)
+  else:
+    external_body.name = ir_data_utils.copy(name)
```
robrussell, line 1177: Why not use `external_body.name = ir_data_utils.copy(name)` in both cases? Or does `CopyFrom()` not overwrite all fields?

robrussell (commented):
Did you intend to also update `header_generator_test.py`? It's the only other place that `CopyFrom` is used.

compiler/front_end/module_ir.py
```diff
@@ -150,7 +150,10 @@ def _really_build_ir(parse_tree, used_productions):
     used_productions.add(parse_tree.production)
     result = _handlers[parse_tree.production](*parsed_children)
     if parse_tree.source_location is not None:
-      result.source_location.CopyFrom(parse_tree.source_location)
+      if result.source_location:
```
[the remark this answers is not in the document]
EricRahm: Yeah, our tests would fail otherwise (as well as the real world tests I run through).

compiler/front_end/module_ir.py
EricRahm: `update` updates the fields of an *existing* `ir_data` A with the set fields of another `ir_data` B (so fields that are already set in A will not be overwritten if they're not set in B) . In the else case there is no `ir_data` to update, so we create one with `copy`.

robrussell: Yeah, I got the logic in this snippet but `update` delegates to `CopyFrom`. I guessed `CopyFrom` might overwrite existing field but it sounds like that's not the case so we need to keep the existing `external_body.name` around.

robrussell (changes requested):
Not sure if you saw the question about `header_generator_test.py` - it has the only remaining direct calls to `CopyFrom`, did you intend to update those too?

EricRahm:
> Not sure if you saw the question about `header_generator_test.py` - it has the only remaining direct calls to `CopyFrom`, did you intend to update those too?

It's used in a few files, the changes I made here were instances that aren't already using a builder. The ones in `header_generator_test` are all essentially `builder(...).CopyFrom`.

robrussell (approved):

robrussell (approved):

PR #146, opened 2024-06-04
Fixes `_make_source_location_tests()` build error.

robrussell:
It looks like there`ir_pb2` was still present in `module_ir_test.py` after #141 . This renames the last couple references to `ir_data`. Tests pass locally with this change.


robrussell:
Addressed in a different PR, reverting.

PR #147, opened 2024-06-04
Runs compilation failure tests.

robrussell:
There are a couple negative compilation tests in `compiler/front_end/module_ir_test.py` named `anonymous bits does not allow documentation` and `anonymous bits does not allow subtypes`. These weren't being run because `_make_negative_tests()`wasn't being called.


robrussell:
@reventlov or @AaronWebster care to take a look?

Issue #154, opened 2024-07-15
Bazel: cannot apply `tags` or `target_compatible_with` on `emboss_cc_library` rules


armandomontanez:
Because `emboss_cc_library` doesn't capture and forward kwargs to the underlying rules, attributes inherent to all rules (like `tags` and `target_compatible_with`) cannot be set on `emboss_cc_library`. Should be a pretty trivial fix.

PR #155, opened 2024-07-15
Forward kwargs to emboss_cc_library

armandomontanez:
Forwards kwargs to emboss_cc_library to fix https://github.com/google/emboss/issues/154.


EricRahm:
@armandomontanez  Can you add a test that exercises this so that we don't break it in the future? @reventlov @robrussell can you take a look?

armandomontanez:
> @armandomontanez Can you add a test that exercises this so that we don't break it in the future? @reventlov @robrussell can you take a look?

Added a trivial instance that exercises this change.

reventlov (approved):

Issue #156, opened 2024-07-16
Extraneous `<ostream>` in generated headers


EricRahm:
There's an extraneous `#include <ostream>` in our generated headers that causes compilation errors on systems where that's not available.

PR #157, opened 2024-07-16
Remove unused <ostream>

EricRahm:
`<ostream>` isn't needed unless enum traits are being used, in that case it will be included by the files that need it. Removing this fixes a failure for targets that don't have `<ostream>` available.

Fixes: #156


EricRahm:
fyi @reventlov @robrussell

reventlov (approved):

PR #158, opened 2024-07-18
Update set of supported alias-safe types

EricRahm:
Renames `IsChar` to `IsAliasSafe`, removes `signed char` as an alias safe type and adds in conditional support for `std::byte` as an anlias safe type if available.

Fixes: #116

### DIFF
--- a/runtime/cpp/emboss_cpp_types.h
+++ b/runtime/cpp/emboss_cpp_types.h
@@ -17,6 +17,7 @@
 #define EMBOSS_RUNTIME_CPP_EMBOSS_CPP_TYPES_H_
 
 #include <climits>
+#include <cstddef>
 #include <cstdint>
 #include <type_traits>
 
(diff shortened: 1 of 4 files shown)


runtime/cpp/emboss_cpp_types.h
```diff
@@ -77,21 +78,23 @@ struct LeastWidthInteger<8> final {
   using Signed = ::std::int8_t;
 };
 
-// IsChar<T>::value is true if T is a character type; i.e. const? volatile?
+// IsAliasSafe<T>::value is true if T is a character type; i.e. const? volatile?
 // (signed|unsigned)? char.
```
reventlov, line 82: Comment is now incorrect.

reventlov (commented):
Thanks for doing this, just one minor nit.

runtime/cpp/emboss_cpp_types.h
EricRahm: Fixed!

reventlov (approved):

PR #163, opened 2024-07-30
Use bzlmod to manage dependencies

tpudlik:
This will also make it easier to maintain emboss in the BCR (https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/emboss).

Ideally we'd also set the `version` field of `module`, but the version value is unknown until the release workflow actually runs (because it encodes the release timestamp). In practice we should probably change the versioning scheme to pull the version number from `MODULE.bazel`, but I imagine that requires some discussion.


jasongraffius:
I'm not too familiar with Bazel modules. Will this change affect anyone users of Emboss that are using Bazel? I suspect not, since this seems to just affect the equivalent of WORKSPACE (?), but I would want to confirm that.

tpudlik:
That's right, existing Bazel projects that pull in Emboss through their WORKSPACE will not be affected.

tpudlik:
Any blockers to merging this?

EricRahm (approved):
lgtm! We'll need to rework #137 but that's fine.

Issue #167, opened 2024-08-07
Emboss fails to run on Python 3.9


jasongraffius:
In addition to #136 , Emboss fails to run on Python 3.9, which unlike 3.8 is not near the end of the support horizon. The failures appear to be related to typechecking:

```python
/home/jason/.cache/bazel/_bazel_jason/3f005b1d454ca9fd185dd4ffbad81e8e/sandbox/linux-sandbox/1366/execroot/com_google_emboss/bazel-out/k8-fastbuild/bin/compiler/back_end/cpp/header_generator_test.runfiles/com_google_emboss/compiler/util/ir_data_fields.py", line 290, in <module>
    def field_specs(obj: IrDataT | type[IrDataT]) -> Mapping[str, FieldSpec]:
TypeError: unsupported operand type(s) for |: 'TypeVar' and 'types.GenericAlias'
```

jasongraffius:
I should mention I have confirmed that python 3.10+ does work, it is only 3.8 and 3.9 of the supported versions that are failing.

EricRahm:
Yeah that's probably my recent changes. Having a check in CI will definitely help prevent this in the future.

PR #169, opened 2024-08-08
Fix for Python 3.9

jasongraffius:
Python 3.9 is still supported, but Python 3.9 does not support type unions with the `|` operator. This change removes these in favor of `Union[...]` or `Optional[...]` as appropriate.

Confirmed all tests pass using Python 3.9


jasongraffius:
This fixes #167

EricRahm (approved):
lgtm, thanks for finding the issue and providing a fix so quickly!

PR #170, opened 2024-08-08
Ensure runners use min supported python version

jasongraffius:
With this change GitHub runners running the verify-pull-request actions will use the python version specified in .python-version


jasongraffius:
Abandoning this in favor of the approach in #137

Issue #152, opened 2024-07-03
Add .gni containing lists of files needed by GN


BenjaminLawson:
Currently we have to maintain lists of Emboss files for use in GN build targets in Pigweed, including the compiler python files and the cpp_utils headers. It would be much more maintainable if there was a .gni file in upstream Emboss contains these file lists.

EricRahm:
I totally get why you want this, I'm just not sure how the emboss project will keep it up to date. We're currently only setup for bazel builds. Maybe adding a tool that can generate these files as part of the bazel build is what we want?

EricRahm:
For concrete examples we have:
1. Pigweed's [build_defs.gni](https://cs.opensource.google/pigweed/pigweed/+/main:third_party/emboss/build_defs.gni;l=105-148;drc=1643306da16c9b2c914eebb087aeb267a7ae139c)
2. Fuchsia's [build_defs.gni](https://osscs.corp.google.com/fuchsia/fuchsia/+/main:third_party/github.com/google/emboss/build_defs.gni;drc=da22575b7d000c3284dec6f1b3f82a3e679a238e)
3. An unlanded but being prototyped version for Android:
```python
python_library_host {
    name: "emboss_compiler",
    srcs: [
        "compiler/back_end/cpp/*.py",
        "compiler/back_end/util/*.py",
        "compiler/front_end/*.py",
        "compiler/util/*.py",
    ],
    data: [
        "compiler/back_end/cpp/generated_code_templates",
        "compiler/front_end/error_examples",
        "compiler/front_end/prelude.emb",
        "compiler/front_end/reserved_words",
    ],
}
```

Additionally each of those has some sort of target for the C++ header library as well, something along the lines of:
```
pw_source_set("cpp_utils") {
  # emboss depends on a separate checkout not included in pigweed, so
  # ignore gn check for this module.
  check_includes = false
  public = [
    "$dir_pw_third_party_emboss/runtime/cpp/emboss_arithmetic.h",
    "$dir_pw_third_party_emboss/runtime/cpp/emboss_arithmetic_all_known_generated.h",
    "$dir_pw_third_party_emboss/runtime/cpp/emboss_arithmetic_maximum_operation_generated.h",
    "$dir_pw_third_party_emboss/runtime/cpp/emboss_array_view.h",
    "$dir_pw_third_party_emboss/runtime/cpp/emboss_bit_util.h",
    "$dir_pw_third_party_emboss/runtime/cpp/emboss_constant_view.h",
[...]

PR #171, opened 2024-08-13
Provide build_info.json file for downstream builds

studgeek:
Currently Emboss only provides build rules for bazel, but downstream projects use a variety of build systems (GN, CMake, Makefile, Soong, etc).  So this change provides a JSON file that lists files used in the build for those downstream builds to use.

https://fxrev.dev/1100339 is one example planned usage of this.

Emboss Issue: https://github.com/google/emboss/issues/152
Pigweed Bug: https://pwbug.dev/359386289


studgeek:
Based on on how we land this will also file a follow-up issue to keep bazel and this new build_info.json in sync. Even without that, this change will make downstream, non-bazel usage much cleaner. If the file list is changed we just need to update it here and have downstream roll, this is much easier than having to sync changing the downstream's hard coded list with emboss's new actual list of files (or in case of Fuchsia having to coordinate Pigweed and Emboss rolls at the same time).

build_info.json
```diff
@@ -0,0 +1,63 @@
+{
+  "comment": "Provides build info for use by downstream build rules.",
+  "emboss_py_files": [
```
reventlov, line 3: Maybe name this something like `emboss_compiler_files`?  `generated_code_templates` and `prelude.emb` are not py files.

studgeek: Yes, better. I also removed the `emboss_` prefix.

build_info.json
```diff
@@ -0,0 +1,63 @@
...
+    "compiler/util/ir_data_utils.py",
+    "compiler/util/ir_util.py",
+    "compiler/util/name_conversion.py",
+    "compiler/util/parser_types.py",
+    "compiler/util/resources.py",
+    "compiler/util/simple_memoizer.py",
+    "compiler/util/traverse_ir.py",
+    "embossc"
+  ],
+  "emboss_h_files": [
```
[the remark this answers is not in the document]
studgeek: Done. I also removed the `emboss_` prefix.

reventlov (approved):

build_info.json
```diff
@@ -0,0 +1,63 @@
+{
+  "comment": "Provides build info for use by downstream build rules.",
```
EricRahm, line 2: Maybe just make this json5 so we can have inline comments?

EricRahm (approved):
Thanks Dave, lgtm as well. Small note on the format but I don't think it's a blocker to landing. Let me know if you want to land as-is and we can push it!

build_info.json
EricRahm: Ah yeah that's true. I imagined this being the input for an *emboss* build step that would generate GN, CMake, Bazel, Soong, etc integration files (as well as plain json), but I realize I didn't actually articulate that here.

AaronWebster:
Thanks for putting this together, @studgeek — and for filing #172 alongside it.

We ended up going with the generated source-of-truth approach from the #172 discussion (the BoringSSL-style one). That landed in #232: `build.json` is now the source of truth, `scripts/build_helpers/generate_build_files.py` emits `gen/sources.{json,bzl,cmake,gni}` for Bazel/CMake/GN/plain JSON, and `scripts/build_helpers/manage_build_json.py --validate` keeps it in sync in CI (`verify-pull-request.yml`).

`gen/sources.json` (`embossc_sources` + `emboss_runtime_cpp_sources`) is the direct successor to the `build_info.json` proposed here, covering the downstream file-list use case from #152 while staying in sync automatically — so #172 is closed as resolved.

Since that supersedes this change, I'm going to close it. Thanks again for driving the original need here!

PR #175, opened 2024-09-05
Add `visibility` to the Bazel rule for the Emboss IR.

reventlov:
This is accomplished by allowing `visibility` to be picked up by `kwargs`.


EricRahm (commented):
lgtm!

EricRahm (approved):

PR #176, opened 2024-09-06
Reformat all source with Black.

reventlov:


jasongraffius (approved):
Thanks for doing this!

jasongraffius:
I believe this also will resolve #76

PR #180, opened 2024-09-09
Fix docstring lints as reported by internal linter.

reventlov:


EricRahm (approved):
I wonder if we can run these checks in ci, I'm not a huge fan of them in the first place but I get that downstream wants them.

reventlov:
> I wonder if we can run these checks in ci, I'm not a huge fan of them in the first place but I get that downstream wants them.

I like the lint that makes sure that docstrings are correctly formatted when present.  I'm not a big fan of requiring a docstring on every single function, especially private functions, but I didn't feel like being picky on this particular PR.

Downstream, the lints are an annoyance, but not a major problem.  It should also be possible to turn them off, but I haven't taken the time to figure out how.

PR #197, opened 2024-10-05
Replace `WhichOneof("x")` with `which_x`. 

reventlov:
This change refactors `OneOfField` so that all fields in a given `oneof`
construct share the same backing attributes on their container class --
`which_{oneof name}`, which holds the (string) name of the
currently-active member of the oneof named `{oneof name}` (or `None` if
no member is active), and `_value_{oneof name}`, which holds the value
of the currently-active member (or `None`).

This avoids looping through field specs in order to do an update or to
figure out which member of a `oneof` is currently active.

Since the `WhichOneof()` method is now a trivial read of a
similarly-named attribute, it can be inlined for a small decrease in
overall code size and without sacrificing readability.

As a result of these changes, the compiler now runs 4.5% faster on my
large test `.emb`.

### DIFF
--- a/compiler/util/ir_data.py
+++ b/compiler/util/ir_data.py
@@ -106,20 +106,6 @@ class Message:
         """Indicates if this class has the given field defined and it is set."""
         return getattr(self, name, None) is not None
 
-    # Non-PEP8 name to mimic the Google Protobuf interface.
-    def WhichOneof(self, oneof_name):  # pylint:disable=invalid-name
-        """Indicates which field has been set for the oneof value.
-
-        Args:
-            oneof_name: the name of the oneof construct to test.
-
-        Returns: the field name, or None if no field has been set.
-        """
-        for field_name, oneof in self.field_specs.oneof_mappings:
-            if oneof == oneof_name and self.HasField(field_name):
-                return field_name
-        return None
-
 
 ################################################################################
 # From here to the end of the file are actual structure definitions.
--- a/compiler/util/ir_data_fields.py
+++ b/compiler/util/ir_data_fields.py
@@ -175,7 +175,6 @@ class FilteredIrFieldSpecs:
         self.all_field_specs = specs
         self.field_specs = tuple(specs.values())
         self.dataclass_field_specs = {k: v for k, v in specs.items() if v.is_dataclass}
-        self.oneof_field_specs = {k: v for k, v in specs.items() if v.is_oneof}
         self.sequence_field_specs = tuple(v for v in specs.values() if v.is_sequence)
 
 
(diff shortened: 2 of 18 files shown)


compiler/util/ir_data.py
```diff
@@ -106,20 +106,6 @@ def HasField(self, name):  # pylint:disable=invalid-name
         """Indicates if this class has the given field defined and it is set."""
         return getattr(self, name, None) is not None
 
-    # Non-PEP8 name to mimic the Google Protobuf interface.
-    def WhichOneof(self, oneof_name):  # pylint:disable=invalid-name
```
EricRahm, line 110: 🥳

compiler/util/ir_data.py
```diff
@@ -106,20 +106,6 @@ def HasField(self, name):  # pylint:disable=invalid-name
...
-        """Indicates which field has been set for the oneof value.
-
-        Args:
-            oneof_name: the name of the oneof construct to test.
-
-        Returns: the field name, or None if no field has been set.
-        """
-        for field_name, oneof in self.field_specs.oneof_mappings:
```
EricRahm, line 118: At this point I believe `oneof_mappings` is only referenced in tests, I think we can remove it from `field_specs` (and any support for it) entirely.

compiler/util/ir_data_fields.py
```diff
@@ -399,34 +399,37 @@ def __init__(self, oneof: str) -> None:
...
+            return None
 
     def __set__(self, obj, value):
         if value is self:
             # This will happen if the dataclass uses the default value, we just
             # default to None.
             value = None
 
-        if value is not None:
-            # Clear the others
```
EricRahm, line 422: I'm slightly concerned not clearing the others will open us up to serialization going wonky. We should double-check the logic there and add a test where we set the oneof multiple times and confirm the correct oneof is set.

EricRahm (commented):
Just one slight concern around serialization, I'm not 100% sure we'll do the right thing there.

compiler/util/ir_data_fields.py
reventlov: The others are implicitly cleared because all of the members in a `oneof` now share the same `which_{oneof_name}` and `_value_{oneof_name}` members.

compiler/util/ir_data.py
reventlov: Good catch.  Done.

EricRahm (approved):
lgtm, one minor cleanup but not a blocker.

Something I was wondering as a follow-up is if using an enum-ish thing would be more performant for the `which_` fields. As in:
```py
if foo.which_type is foo.INTEGER:
```
vs
```py
if foo.which_type == "integer":
```
Gut feeling is this is super diminishing returns, but just kind of curious.

compiler/util/ir_data_fields.py
```diff
@@ -177,9 +177,6 @@ def __init__(self, specs: Mapping[str, FieldSpec]):
         self.dataclass_field_specs = {k: v for k, v in specs.items() if v.is_dataclass}
         self.oneof_field_specs = {k: v for k, v in specs.items() if v.is_oneof}
         self.sequence_field_specs = tuple(v for v in specs.values() if v.is_sequence)
-        self.oneof_mappings = tuple(
-            (k, v.oneof) for k, v in self.oneof_field_specs.items() if v.oneof
```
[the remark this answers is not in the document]
reventlov: Done.

PR #201, opened 2024-10-10
Error example generator

reventlov:
This is a program that generates examples of erroneous parses, along
with some information about the potential errors, with the intention of
allowing developers to quickly find out what states do *not* have
associated error messages.

Ideally, this program would be used to check that there *are no* states
that are missing error messages, but actually adding all of those
messages is not feasible right now: at this change,
enumerate_parse_errors finds 5808 states that are missing messages.

### DIFF
--- a/compiler/front_end/BUILD
+++ b/compiler/front_end/BUILD
@@ -420,6 +420,16 @@ py_test(
 )
 
 py_binary(
+    name = "enumerate_parse_errors",
+    srcs = ["enumerate_parse_errors.py"],
+    deps = [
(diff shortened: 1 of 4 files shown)


reventlov:
This is an update of a PR from some time in 2023 which I never actually finished, which, itself, is an update of an internal CL from 2015(!) that was approved but never actually checked in, so... it may be a little bit disjointed.

It's also pretty algorithmically subtle.  I have tried to document it, but I am not confident that I was able to make it clear.

PR #204, opened 2024-10-11
Emit a warning when the cached parser is not used.

reventlov:
This change checks to see if the cached parser was discarded due to a mismatch between the cached parser and the grammar specified in module_ir.py, and, if so, emits a warning that the cached parser was not used, along with informational messages on the nature of the mismatch.

Adjusted the "warning" color from magenta to yellow.  (This is the first warning in Emboss, so no magenta messages would have ever been emitted.)

Adjusted the "note" color from "bright black" (dark grey) to "white" (light grey), becaused at least some terminals display "bright black" as just black.

### DIFF
--- a/compiler/front_end/emboss_front_end.py
+++ b/compiler/front_end/emboss_front_end.py
@@ -29,6 +29,7 @@ import sys
 
 from compiler.front_end import glue
 from compiler.front_end import module_ir
+from compiler.front_end import parser
 from compiler.util import error
 from compiler.util import ir_data_utils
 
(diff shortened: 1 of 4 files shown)


reventlov:
An example of the warning, as displayed on my terminal:
parser_cache_warning

compiler/front_end/emboss_front_end.py
```diff
@@ -151,6 +152,34 @@ def _find_and_read(file_name):
...
+                        None,
+                        "Cached parser does not match actual grammar; using newly-generated parser.",
+                    )
+                ]
+                + extra_production_notes
+                + missing_production_notes
+            ],
+            None,
+            color_output,
+        )
```
jasongraffius, line 180: Having the specific extra/missing productions is perfect, this will also be very helpful when debugging a grammar change

jasongraffius (approved):
LGTM

PR #205, opened 2024-10-14
Convert Location to a namedtuple, and associated cleanup

reventlov:
This change makes the `Location` dataclass, which does not change
frequently, into a new `SourceLocation` namedtuple, and changes the
`SourceLocation` serialization.  As a result, with this change:

*   `embossc` runs about 25% faster on a large (7kLOC) input; `python3
    -OO embossc` runs about 19% faster on the same input.
*   Serialized IR is about 45% smaller.

Details:

*   Replace the `ir_data.Location` dataclass with a new
    `parser_types.SourceLocation` namedtuple.  The rename helps clarify
    the difference between a location within source code
    (`SourceLocation`) and a location within a structure
    (`FieldLocation`).
*   Similarly, replace `ir_data.Position` with
    `parser_types.SourcePosition`.
*   Update any place that edits a `SourceLocation` with an appropriate
    assignment; e.g., `x.source_location.end = y` becomes
    `x.source_location = x.source_location._replace(end=y)`.  In most
    cases, several fields were updated consecutively; those updates are
    been merged.
*   Update the JSON serialization to use the compact format.
*   Replace `format_location()` and `format_position()` with
    `__str__()` methods on `SourceLocation` and `SourcePosition`,
    respectively.
*   Replace `parse_location()` and `parse_position()` with `from_str()`
    class methods on `SourceLocation` and `SourcePosition`,
    respectively.
*   Move the `make_location()` functionality into
    `SourceLocation.__new__()`.
*   Update `_to_dict` and `_from_dict` in `IrDataSerializer` to
    stringify and destringify `SourceLocation`.  It is tempting to
    try to do this during the JSON serialization step (with a `default=`
    parameter to `json.dumps` and an `object_hook=` parameter to
    `json.loads`), but it is tricky to get the `object_hook` to know
    when to convert.
*   Centralize the logic for merging `source_location`s into
    `merge_source_locations()`.

### DIFF
--- a/compiler/front_end/module_ir.py
+++ b/compiler/front_end/module_ir.py
@@ -40,7 +40,7 @@ class _List(object):
     def __init__(self, l):
         assert isinstance(l, list), "_List object must wrap list, not '%r'" % l
         self.list = l
-        self.source_location = ir_data.Location()
+        self.source_location = parser_types.SourceLocation()
 
 
 class _ExpressionTail(object):
@@ -65,7 +65,7 @@ class _ExpressionTail(object):
     def __init__(self, operator, expression):
         self.operator = operator
         self.expression = expression
-        self.source_location = ir_data.Location()
+        self.source_location = parser_types.SourceLocation()
 
 
 class _FieldWithType(object):
@@ -76,7 +76,7 @@ class _FieldWithType(object):
     def __init__(self, field, subtypes=None):
         self.field = field
         self.subtypes = subtypes or []
-        self.source_location = ir_data.Location()
+        self.source_location = parser_types.SourceLocation()
 
 
 def build_ir(parse_tree, used_productions=None):
@@ -153,11 +153,11 @@ def _really_build_ir(parse_tree, used_productions):
         ]
         used_productions.add(parse_tree.production)
         result = _handlers[parse_tree.production](*parsed_children)
-        if parse_tree.source_location is not None:
-            if result.source_location:
-                ir_data_utils.update(result.source_location, parse_tree.source_location)
+        if parse_tree.source_location:
+            if isinstance(result, tuple):
+                result = result._replace(source_location=parse_tree.source_location)
             else:
-                result.source_location = ir_data_utils.copy(parse_tree.source_location)
+                result.source_location = parse_tree.source_location
         return result
     else:
         # For leaf nodes, the temporary "IR" is just the token.  Higher-level rules
@@ -191,7 +191,7 @@ def _handles(production_text):
 
 def _make_prelude_import(position):
     """Helper function to construct a synthetic ir_data.Import for the prelude."""
-    location = parser_types.make_location(position, position)
+    location = parser_types.SourceLocation(position, position)
     return ir_data.Import(
         file_name=ir_data.String(text="", source_location=location),
         local_name=ir_data.Word(text="", source_location=location),
@@ -279,7 +279,7 @@ def _file(leading_newlines, docs, imports, attributes, type_definitions):
         and not attributes.list
         and not type_definitions.list
     ):
-        module_source_location = parser_types.make_location((1, 1), (1, 1))
+        module_source_location = parser_types.SourceLocation((1, 1), (1, 1))
     else:
         module_source_location = None
 
@@ -462,10 +462,10 @@ def _expression(expression):
 )
 def _choice_expression(condition, question, if_true, colon, if_false):
     """Constructs an IR node for a choice operator (`?:`) expression."""
-    location = parser_types.make_location(
+    location = parser_types.SourceLocation(
         condition.source_location.start, if_false.source_location.end
     )
-    operator_location = parser_types.make_location(
+    operator_location = parser_types.SourceLocation(
         question.source_location.start, colon.source_location.end
     )
     # The function_name is a bit weird, but should suffice for any error
@@ -490,7 +490,7 @@ def _no_op_comparative_expression(expression):
     "    additive-expression inequality-operator additive-expression"
 )
 def _comparative_expression(left, operator, right):
-    location = parser_types.make_location(
+    location = parser_types.SourceLocation(
         left.source_location.start, right.source_location.end
     )
     return ir_data.Expression(
@@ -538,7 +538,7 @@ def _binary_operator_expression(expression, expression_right):
     """
     e = expression
     for right in expression_right.list:
-        location = parser_types.make_location(
+        location = parser_types.SourceLocation(
             e.source_location.start, right.source_location.end
         )
         e = ir_data.Expression(
@@ -611,7 +611,7 @@ def _chained_comparison_expression(expression, expression_right):
     comparisons = []
     for i in range(0, len(sequence) - 1, 2):
         left, operator, right = sequence[i : i + 3]
-        location = parser_types.make_location(
+        location = parser_types.SourceLocation(
             left.source_location.start, right.source_location.end
         )
         comparisons.append(
@@ -627,7 +627,7 @@ def _chained_comparison_expression(expression, expression_right):
         )
     e = comparisons[0]
     for comparison in comparisons[1:]:
-        location = parser_types.make_location(
+        location = parser_types.SourceLocation(
             e.source_location.start, comparison.source_location.end
         )
         e = ir_data.Expression(
@@ -717,7 +717,7 @@ def _expression_right_production(operator, expression):
 # allowed, but "+-5" or "-+-something" are not.
 @_handles("negation-expression -> additive-operator bottom-expression")
 def _negation_expression_with_operator(operator, expression):
-    phantom_zero_location = ir_data.Location(
+    phantom_zero_location = parser_types.SourceLocation(
         start=operator.source_location.start, end=operator.source_location.start
     )
     return ir_data.Expression(
@@ -733,7 +733,7 @@ def _negation_expression_with_operator(operator, expression):
                 expression,
             ],
             function_name=operator,
-            source_location=ir_data.Location(
+            source_location=parser_types.SourceLocation(
                 start=operator.source_location.start, end=expression.source_location.end
             ),
         )
@@ -759,7 +759,7 @@ def _bottom_expression_function(function, open_paren, arguments, close_paren):
             function=_text_to_function(function.text),
             args=arguments.list,
             function_name=function,
-            source_location=ir_data.Location(
+            source_location=parser_types.SourceLocation(
                 start=function.source_location.start,
                 end=close_paren.source_location.end,
             ),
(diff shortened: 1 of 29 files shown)


reventlov:
The +/- on this PR shows a lot of lines removed, but they're mostly in IR fragments -- the actual code changes are roughly a wash in LOC.

compiler/front_end/module_ir_test.py
```diff
@@ -4025,29 +3956,25 @@ def _check_source_location(source_location, path, min_start, max_end):
...
                         path, path, start.line, end.line
                     )
                 )
             elif start.line == end.line:
-                if (
-                    start.HasField("column")
-                    and end.HasField("column")
-                    and start.column > end.column
-                ):
+                if start.column and end.column and start.column > end.column:
```
EricRahm, line 3977: nit: we shouldn't have the `start.column and end.column` portion here, they'll always be present, this is just checking that they're not zero now right?

compiler/front_end/module_ir_test.py
```diff
@@ -4025,29 +3956,25 @@ def _check_source_location(source_location, path, min_start, max_end):
...
         result.append("{}.start missing".format(path))
     else:
         start = source_location.start
-    if not source_location.HasField("end"):
+    if not source_location.end:
         result.append("{}.end missing".format(path))
     else:
         end = source_location.end
 
     if start and end:
```
EricRahm, line 3968: I'm not sure this is checking what we want now? Maybe `is not None` is more corect.

compiler/front_end/module_ir_test.py
```diff
@@ -4025,29 +3956,25 @@ def _check_source_location(source_location, path, min_start, max_end):
     result = []
     start = None
     end = None
-    if not source_location.HasField("start"):
+    if not source_location.start:
```
EricRahm, line 3959: `start` is always present now, but the __bool__ impl makes this check if it's a defaulted value, is that what we want? Should we maybe just add a `is_default()` method instead?

compiler/front_end/module_ir_test.py
```diff
@@ -4025,29 +3956,25 @@ def _check_source_location(source_location, path, min_start, max_end):
...
         result.append("{}.start missing".format(path))
     else:
         start = source_location.start
-    if not source_location.HasField("end"):
+    if not source_location.end:
```
EricRahm, line 3963: same comment

compiler/front_end/module_ir_test.py
```diff
@@ -4057,12 +3984,12 @@ def _check_source_location(source_location, path, min_start, max_end):
     for name, field in (("start", start), ("end", end)):
         if not field:
```
EricRahm, line 3985: similar comment as above, should this be `field is None`

compiler/front_end/module_ir_test.py
```diff
@@ -4057,12 +3984,12 @@ def _check_source_location(source_location, path, min_start, max_end):
...
             continue
-        if field.HasField("line"):
+        if field.line:
```
EricRahm, line 3987: ...and again

compiler/front_end/module_ir_test.py
```diff
@@ -4057,12 +3984,12 @@ def _check_source_location(source_location, path, min_start, max_end):
...
             if field.line <= 0:
                 result.append("{}.{}.line <= 0 ({})".format(path, name, field.line))
         else:
             result.append("{}.{}.line missing".format(path, name))
-        if field.HasField("column"):
+        if field.column:
```
EricRahm, line 3992: ... and again

compiler/util/parser_types.py
```diff
@@ -12,72 +12,139 @@
...
-        start.line == end.line and start.column > end.column
+    @staticmethod
+    def from_str(value):
+        try:
+            l, c = value.split(":")
+            return SourcePosition(line=int(l.strip()), column=int(c.strip()))
+        except Exception as e:
+            raise ValueError(f"{repr(value)} is not a valid SourcePosition.")
+
+    def __bool__(self):
```
EricRahm, line 54: I'm not a 100% sure adding a bool operator is a good idea or not, see my notes above. A `is_default()` might be more clear about what our intentions are.

compiler/util/parser_types.py
```diff
@@ -12,72 +12,139 @@
...
-    return ir_data.Location(start=start, end=end, is_synthetic=is_synthetic)
 
+    def __str__(self):
+        suffix = ""
+        if self.is_disjoint_from_parent:
+            suffix += "^"
+        if self.is_synthetic:
+            suffix += "*"
+        return (
+            f"{self.start or SourcePosition()}-{self.end or SourcePosition()}{suffix}"
```
EricRahm, line 99: The `or SourcePosition()` shouldn't be necessary here right?

compiler/util/parser_types.py
```diff
@@ -12,72 +12,139 @@
...
+            return SourceLocation(
+                start=SourcePosition.from_str(start),
+                end=SourcePosition.from_str(end),
+                is_synthetic=is_synthetic,
+                is_disjoint_from_parent=is_disjoint_from_parent,
+            )
+        except Exception as e:
+            raise ValueError(f"{repr(original_value)} is not a valid SourceLocation.")
+
+    def __bool__(self):
```
EricRahm, line 124: Same thoughts on bool here.

EricRahm (commented):
This is great, thanks for doing this. A few questions around handing of default values

compiler/front_end/module_ir.py
```diff
@@ -153,11 +153,11 @@ def _really_build_ir(parse_tree, used_productions):
...
         result = _handlers[parse_tree.production](*parsed_children)
-        if parse_tree.source_location is not None:
-            if result.source_location:
-                ir_data_utils.update(result.source_location, parse_tree.source_location)
+        if parse_tree.source_location:
+            if isinstance(result, tuple):
+                result = result._replace(source_location=parse_tree.source_location)
             else:
-                result.source_location = ir_data_utils.copy(parse_tree.source_location)
+                result.source_location = parse_tree.source_location
```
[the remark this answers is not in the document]
reventlov: Yeah, more or less the point of this change is to make it so that it is never necessary to deep copy `SourceLocation` -- it's frozen, so it doesn't matter if multiple nodes point to the same `SourceLocation` instance.  About half of the time in `_copy()` was just copying `Location`.

It would be (relatively) cheap to special case `update` and `copy` to check if their arguments are `tuple` (specifically checking for anything created by `namedtuple()` is harder), but I don't really see the point?  Right now, they just raise an exception if you give them something that isn't an IR dataclass, so any mistakes will be caught quickly.

compiler/util/parser_types.py
reventlov: Done.

Left over from a previous iteration, when I allowed `start` and `end` to be `None`.

compiler/front_end/module_ir_test.py
```diff
@@ -4025,29 +3956,25 @@ def _check_source_location(source_location, path, min_start, max_end):
...
         start = source_location.start
-    if not source_location.HasField("end"):
+    if not source_location.end:
         result.append("{}.end missing".format(path))
     else:
         end = source_location.end
 
     if start and end:
-        if start.HasField("line") and end.HasField("line"):
+        if start.line and end.line:
```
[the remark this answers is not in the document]
reventlov: I tightened up the code in here, mostly to take advantage of the `tuple` ordering.  I think that the altered version addresses all of your comments.

This function is used to check the `source_location` invariants after `module_ir.build_ir()` is done -- basically, every node with a `source_location` field must have a source location, the source location must be inside the parent's source location, and both `line` and `column` must be `>= 1`.

I also added a number of tests to `parser_types_test` to cover the internal invariants in `SourcePosition` and `SourceLocation`.

compiler/front_end/module_ir_test.py
reventlov: The case where `not field and field is not None` is covered a few lines above, in the `if not source_location.start`/`.end`.

compiler/front_end/module_ir_test.py
reventlov: This is mostly a matter of taste.  The `__bool__` method seems more Pythonic to me (truthy/falsy values rather than explicit tests), but your preference should probably override here.

reventlov: One practical effect of using `__bool__` is that you can do:

```Python
if x.source_location:
    # ...
```

instead of:

```Python
if x.source_location is not None and not x.source_location.is_default():
    # ...
```

compiler/front_end/module_ir.py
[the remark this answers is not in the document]
EricRahm: Yeah I think it's fine as-is, it just took me a second to grok why we weren't using the standard methods for this anymore.

compiler/front_end/module_ir_test.py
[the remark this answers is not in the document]
EricRahm: Ah okay, so for a source location to be valid it must not be: (0,0), (0,1), (1,0) ?

compiler/front_end/module_ir_test.py
EricRahm: Oh right I see, we have an overall "end is missing"

compiler/util/parser_types.py
```diff
@@ -12,72 +12,137 @@
...
+    "PositionTuple", ["line", "column"], defaults=[0, 0]
+)
 
-def _parse_position(text):
-    """Parses an ir_data.Position from "line:column" (e.g., "1:2")."""
-    line, column = text.split(":")
-    return _make_position(int(line), int(column))
 
+class SourcePosition(PositionTuple):
+    """A zero-width position within a source file."""
```
EricRahm, line 34: Perhaps document that `(0,0)` is the default position, and is considered invalid/falsey/unset.

EricRahm (approved):
lgtm as long as `(0,0)` is always invalid things make sense to me.

compiler/util/parser_types.py
reventlov: Done.

compiler/front_end/module_ir_test.py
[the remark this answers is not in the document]
reventlov: At least in here, yes.  `(0, 0)` is "valid" in the sense that you might have a node that didn't "come from" something in the source text, but `module_ir` shouldn't be creating any of those.

I just updated `SourcePosition.__new__` to `assert`-fail on (0, x) and (x, 0), and `SourceLocation.__new__` to `assert`-fail if `start and not end` or `end and not start`, and removed the checks here for only `line` or `column` being `0`.

reventlov:
> Gotcha, so if `(0,0)` is always invalid I agree the bool operator is fine, I think I just didn't get that

It's invalid in the sense that it does not refer to an actual location (using 1-based numbering, not 0-based, because that's how lines and columns are customarily numbered).  I added some explanation to the docstring.

compiler/util/parser_types.py
EricRahm: We cleared this up, this is fine.

compiler/front_end/module_ir_test.py
EricRahm: We cleared this up, this is fine.

compiler/util/parser_types.py
EricRahm: This is fine.

EricRahm (approved):
lgtm!

PR #211, opened 2024-11-14
Add test profiles that disable all `EMBOSS_CHECK`s.

reventlov:
This catches cases where C++ code inadvertently includes side effects inside of an `EMBOSS_CHECK`; the default `EMBOSS_CHECK` uses `assert`, which will be omitted when `NDEBUG` is defined.

Also remove the `EMBOSS_FORCE_ALL_CHECKS` define, which is a holdover from the glog-based `EMBOSS_DCHECK` implementation in the original, Google-internal implementation.

### DIFF
--- a/compiler/back_end/cpp/build_defs.bzl
+++ b/compiler/back_end/cpp/build_defs.bzl
@@ -20,14 +20,26 @@ def emboss_cc_test(name, copts = None, no_w_sign_compare = False, **kwargs):
     """Generates cc_test rules with and without -DEMBOSS_NO_OPTIMIZATIONS."""
     native.cc_test(
         name = name,
-        copts = ["-DEMBOSS_FORCE_ALL_CHECKS"] + (copts or []),
+        copts = copts or [],
         **kwargs
     )
     native.cc_test(
         name = name + "_no_opts",
         copts = [
             "-DEMBOSS_NO_OPTIMIZATIONS",
-            "-DEMBOSS_FORCE_ALL_CHECKS",
+        ] + ([] if no_w_sign_compare else ["-Wsign-compare"]) + (copts or []),
+        **kwargs
+    )
+    native.cc_test(
+        name = name + "_no_checks",
+        copts = ["-DEMBOSS_SKIP_CHECKS"] + (copts or []),
+        **kwargs
+    )
+    native.cc_test(
+        name = name + "_no_checks_no_opts",
+        copts = [
+            "-DEMBOSS_NO_OPTIMIZATIONS",
+            "-DEMBOSS_SKIP_CHECKS",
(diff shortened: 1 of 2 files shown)


compiler/back_end/cpp/build_defs.bzl
```diff
@@ -20,14 +20,26 @@ def emboss_cc_test(name, copts = None, no_w_sign_compare = False, **kwargs):
...
         name = name + "_no_opts",
         copts = [
             "-DEMBOSS_NO_OPTIMIZATIONS",
-            "-DEMBOSS_FORCE_ALL_CHECKS",
+        ] + ([] if no_w_sign_compare else ["-Wsign-compare"]) + (copts or []),
+        **kwargs
+    )
+    native.cc_test(
+        name = name + "_no_checks",
+        copts = ["-DEMBOSS_SKIP_CHECKS"] + (copts or []),
```
EricRahm, line 35: Are we intentionally skipping sign compare here because this is "no checks"?

runtime/cpp/emboss_defines.h
```diff
@@ -66,9 +66,17 @@
 //
 // By default, checks are only enabled on non-NDEBUG builds.  (Note that all
 // translation units MUST be built with the same value of NDEBUG!)
+//
+// The EMBOSS_SKIP_CHECKS parameter allows all CHECKs to be compiled out
+// without -DNDEBUG; this is useful for testing.
 #if !defined(EMBOSS_CHECK)
+#if EMBOSS_SKIP_CHECKS
```
EricRahm, line 73: Should this be `#if defined(EMBOSS_SKIP_CHECKS)`? I'm a little rusty on preprocessor rules

EricRahm (approved):
Two minor comments, overall concepts makes sense. Thanks for adding this!

compiler/back_end/cpp/build_defs.bzl
reventlov: I should figure out why we aren't checking `-Wsign-compare` on the `opts` tests, but that is orthogonal to this change.

runtime/cpp/emboss_defines.h
reventlov: It probably should be for consistency with surrounding code.

Technically, they're almost identical -- `#if SYMBOL` is false if `SYMBOL` is undefined or evaluates to 0; `#ifdef SYMBOL` or `#if defined(SYMBOL)` is only false if `SYMBOL` is undefined.  Passing `-DSYMBOL` implicitly `#defines` `SYMBOL` to be `1`.

EricRahm: Ah thanks for the refresher!

PR #212, opened 2024-12-05
Fix tests under `bazel test -c opt`.

reventlov:
With certain versions of GCC, `bazel test -c opt ...` failed, apparently due to a compiler bug.

This change:

1.  Adds a workaround for the GCC bug.
2.  Enables the workaround on versions of GCC known to have the bug.
3.  Documents the bug (in comments).
4.  Enables testing with `-c opt` as part of the commit hook.

Note that the versions covered by #2 are a little wider than may be technically necessary.  As of the time of writing, this affects all released versions of GCC from 12.1 through 14.1, but the GCC fix has been backported to the 12.X and 13.X branches.  It may be possible to reduce the workaround's scope once those are released.

### DIFF
--- a/.github/workflows/verify-pull-request.yml
+++ b/.github/workflows/verify-pull-request.yml
@@ -8,6 +8,7 @@ jobs:
     strategy:
       matrix:
         cpp-compiler: ["clang", "gcc"]
+        options: ["-c opt", ""]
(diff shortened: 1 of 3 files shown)


jasongraffius (approved):
LGTM

PR #214, opened 2024-12-13
Add error message for obscure case with parameters.

reventlov:
Prior to this change, if a user omitted the length of an integer parameter *and* used that parameter as an argument to an operator or function, the compiler with throw an `AttributeError` instead of providing a useful message.

This is because the check that integer runtime parameters have an explicit size happened later than the function in `expression_bounds.py` that tried to use the size.

This change moves the check earlier (incidentally spliting `check_constraints` into `check_constraints` and
`check_early_constraints`), and also gives it a better error message.

### DIFF
--- a/compiler/front_end/glue.py
+++ b/compiler/front_end/glue.py
@@ -319,6 +319,7 @@ def process_ir(ir, stop_before_step):
         symbol_resolver.resolve_field_references,
         type_check.annotate_types,
         type_check.check_types,
+        constraints.check_early_constraints,
         expression_bounds.compute_constants,
         attribute_checker.normalize_and_verify,
         constraints.check_constraints,
--- a/compiler/front_end/constraints.py
+++ b/compiler/front_end/constraints.py
@@ -713,6 +713,7 @@ def _check_bounds_on_runtime_integer_expressions(
 def _attribute_in_attribute_action(a):
     return {"in_attribute": a}
 
+
 def check_early_constraints(ir):
     errors = []
     traverse_ir.fast_traverse_ir_top_down(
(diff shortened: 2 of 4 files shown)


EricRahm:
Taking a look now!

compiler/front_end/glue.py
```diff
@@ -319,6 +319,7 @@ def process_ir(ir, stop_before_step):
         symbol_resolver.resolve_field_references,
         type_check.annotate_types,
         type_check.check_types,
+        constraints.check_early_constraints,
```
EricRahm, line 322: Out of curiosity does this additional pass add much overhead to the end-to-end codegen time?

EricRahm (approved):
Thanks for adding this check, lgtm. Just a very minor concern about whether or not an additional pass will add more overhead (my guess is it's negligible).

compiler/front_end/glue.py
reventlov:
```
> python3 -m cProfile ./embossc --output-path ~/scratch ~/scratch/large.emb

         25766374 function calls (23764959 primitive calls) in 17.613 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    181/1    0.014    0.000   18.280   18.280 {built-in method builtins.exec}
[...]
        1    0.000    0.000    0.003    0.003 constraints.py:717(check_early_constraints)
[...]
        8    0.000    0.000    0.002    0.000 constraints.py:365(_check_type_requirements_for_parameter_type)
[...]
```

Looks like <0.02% of the overall time compilation.  `large.emb` does have a few parameters, but is not parameter-heavy; I imagine that an `.emb` with a lot of parameters would pay a correspondingly higher compile time cost.

PR #215, opened 2024-12-19
Handle very large `.emb` files.

reventlov:
This change switches parse tree handling to use iteration (with an explicit stack) instead of recursion, which:

*   Allows large (>~1000 entity) `.emb` files to be formatted.
*   Allows very large (>~16k entity) `.emb` files to be compiled.

The difference in sizes in the previous code was due to `module_ir.py` hackily increasing the recursion limit: while this more or less worked, it was a little dangerous (it ran the risk of blowing out the C stack, depending on platform) and only increased the limit.  This change removes the limit entirely (at least, up to the available memory on the system).


jasongraffius (approved):
LGTM

PR #216, opened 2024-12-19
Add `alignas` specifier to buffer in explicit alignment test.

reventlov:
This fixes the `clang` test failures.

Also update workflow to emit test logs on failure.


EricRahm (approved):
Oh nice catch!

PR #219, opened 2025-01-30
Fix ASAN failure in `array.rend()`.

reventlov:
The iterator constructor used `array[index]` to initialize an internal cached view (necessary so that `operator->` can return a pointer), which caused pointer arithmetic overflow when `index` was -1 -- as in the case of `rend()`.

This change uses a checked `array.at(index)` method to get a null view when `index` is out of bounds.

I considered changing `array[index]` to return a null view for out-of-bounds indexes, but user code may be inadvertently relying on that behavior, so I am leaving it out of this bug fix.


EricRahm (approved):
lgtm!

PR #228, opened 2025-08-20
Adds standalone driver `emboss-format` for `compiler/front_end/format.py`

reventlov:
This should both help visibility for the formatter and make it easier for non-Bazel users to use it.


reventlov:
I updated my CLA to add my GitHub username, so hopefully the CLA check will pass now.

jasongraffius:
@reventlov I re-ran the CLA scan, looks like your Github username is validating now in the CLA, but not the email in the author line of the commits.

reventlov:
@jasongraffius I think I have gotten that email address (ben-emboss@xn13.com) in the right place to make CLA bot happy now.

emboss-format
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+
+# Copyright 2025 Benjamin Olmstead
```
jasongraffius, line 3: Should this be Google LLC?

emboss-format
```diff
@@ -0,0 +1,34 @@
...
+import argparse
+import os
+import sys
+
+
+def main(argv):
+    base_path = os.path.dirname(__file__) or "."
+    sys.path.append(base_path)
+
+    from compiler.front_end import format  # pylint:disable=import-outside-toplevel
```
jasongraffius, line 28: Why import inside main? I know we do this in `embossc` as well, but I wasn't sure the purpose of this.

jasongraffius (commented):
Thanks for this! I have a couple of quick questions

emboss-format
reventlov: I don't believe so, since Google does not own the copyright to my work.  The CLA gives Google a very broad license, but is not copyright assignation.

emboss-format
reventlov: Updated.  Here, I was just copying `embossc`.

When I wrote `embossc`, I thought it was necessary to add the script's directory to Python's module path before doing the import, but it looks like [a standard Python installation will put the script's directory into the module search path](https://docs.python.org/3/library/sys_path_init.html).  Either I was testing `embossc` on a weird installation or I got confused somewhere -- I'd guess the latter 🙂.  Either way, I can send a followup PR to fix that script.

jasongraffius (approved):

PR #231, opened 2025-09-30
embossc: Revert path manipulation

jasongraffius:
Restore the sys.path manipulation temporarily while investigating a possible downstream issue.


EricRahm (approved):
lgtm, thanks for the quick revert

PR #234, opened 2025-11-25
Optimize C++ Backend for Conditional Fields    

AaronWebster:
This change significantly optimizes the generated C++ code for
structures utilizing conditional fields.

Key optimizations include:
- Two-Pass Generation: Implemented a strategy to identify and cache
  frequently used subexpressions (e.g., tag reads) within Ok(),
  Equals(), and UncheckedEquals() methods.
  - This involves a first pass to count subexpression usage, followed
    by a second pass that caches repeated expressions in local
    variables while inlining unique ones.
- Reduced Redundancy: Minimizes redundant computations, leading to CPU
  performance improvements and reduced stack usage for unique
  conditions.

Performance Impact (on testdata/many_conditionals.emb with 10,000
iterations x 100 tags):
- Runtime Speedup: ~8.7x faster (from ~1.83s to ~0.21s).
- Binary Size Reduction: ~47% smaller (from 57848 bytes to 30472
  bytes for many_conditionals_benchmark with -c opt).

Additional updates:
- Test infrastructure was enhanced to support passing compiler flags to
  golden tests, resolving an issue with no_enum_traits_golden_test.
- New test data (testdata/many_conditionals.emb) and a benchmark
  (compiler/back_end/cpp/testcode/many_conditionals_benchmark.cc)
  were added to verify performance.

### DIFF
--- a/compiler/back_end/cpp/testcode/enum_case_test.cc
+++ b/compiler/back_end/cpp/testcode/enum_case_test.cc
@@ -16,7 +16,11 @@
 // these tests are regarding the name of enum members, it is likely that if this
 // test would fail, it may fail to compile.
 
+#include <cstdint>
+#include <array>
+
 #include "gtest/gtest.h"
+#include "runtime/cpp/emboss_text_util.h"
 #include "testdata/enum_case.emb.h"
 
 namespace emboss {
--- a/compiler/back_end/cpp/testcode/importer2_test.cc
+++ b/compiler/back_end/cpp/testcode/importer2_test.cc
@@ -14,9 +14,7 @@
 
 // Tests for using types that are imported from imports.
 
-#include <stdint.h>
-
-#include <vector>
+#include <cstdint>
 
 #include "gtest/gtest.h"
 #include "testdata/importer2.emb.h"
--- a/compiler/back_end/cpp/testcode/no_enum_traits_test.cc
+++ b/compiler/back_end/cpp/testcode/no_enum_traits_test.cc
@@ -18,7 +18,6 @@
 
 #include <vector>
 
-#include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "testdata/no_enum_traits.emb.h"
 
--- a/compiler/front_end/BUILD
+++ b/compiler/front_end/BUILD
@@ -457,6 +457,16 @@ py_test(
 )
 
 py_binary(
+    name = "enumerate_parse_errors",
+    srcs = ["enumerate_parse_errors.py"],
+    deps = [
+        ":lr1",
+        ":parser",
+        ":tokenizer",
+    ],
+)
+
+py_binary(
     name = "generate_grammar_md",
     srcs = ["generate_grammar_md.py"],
     python_version = "PY3",
--- a/compiler/front_end/enumerate_parse_errors.py
+++ b/compiler/front_end/enumerate_parse_errors.py
@@ -1,0 +1,487 @@
+"""Program to generate a list of parse error examples for Emboss.
+
+This is a development aid; when new grammar productions are added, this tool
+can run to generate new error examples.
+
+This program's output should be exhaustive: it exploits knowledge of the
+tokenizer (in particular, what tokens can possibly follow other tokens) and
+deep inspection of the parser tables to quickly find traversals of the parser
+state space that correspond to realizable sequences of tokens.
+"""
+
+from __future__ import print_function
+
+import argparse
+import collections
+import pkgutil
+import sys
+
+from compiler.front_end import lr1
+from compiler.front_end import parser
+from compiler.front_end import tokenizer
+
+
+def _parse_command_line(argv):
+    """Parses the given command-line arguments."""
+    parser = argparse.ArgumentParser(
+        description="Generator of examples inputs for Emboss parser errors.",
+        prog=argv[0],
+    )
+    parser.add_argument(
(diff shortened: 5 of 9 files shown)


compiler/back_end/cpp/generated_code_templates
```diff
@@ -391,11 +392,14 @@ ${write_fields}
 // ** ok_method_test ** ////////////////////////////////////////////////////////
     // If we don't have enough information to determine whether ${field} is
     // present in the structure, then structure.Ok() should be false.
-    if (!has_${field}.Known()) return false;
-    // If ${field} is present, but not Ok(), then structure.Ok() should be
-    // false.  If ${field} is not present, it does not matter whether it is
-    // Ok().
-    if (has_${field}.ValueOrDefault() && !${field}.Ok()) return false;
+    {
+      const auto emboss_reserved_local_field_present = ${existence_condition};
```
EricRahm, line 396: So I *think* this is intended to inline the contents of `has_<field>`, is that correct? And the goal of that is to reuse a cached instance of a field that is referenced in multiple `has_<field>` methods? Looking at the generated code now it's...pretty verbose. I'm actually a little surprised this would reduce code size.

testdata/golden_cpp/condition.emb.h
```diff
@@ -4265,28 +4386,48 @@ class GenericCorrectNestedConditionalView final {
   bool Ok() const {
     if (!IsComplete()) return false;
 
-    if (!has_x().Known()) return false;
-    if (has_x().ValueOrDefault() && !x().Ok()) return false;
+    const auto emboss_reserved_local_ok_subexpr_1 = x();
```
EricRahm, line 4389: I guess this is an example of what we're trying to optimize.

testdata/golden_cpp/condition.emb.h
```diff
@@ -4265,28 +4386,48 @@ class GenericCorrectNestedConditionalView final {
...
 
+    {
+      const auto emboss_reserved_local_field_present = ::emboss::support::Maybe</**/bool>(true);
+      if (!emboss_reserved_local_field_present.Known()) return false;
+      if (emboss_reserved_local_field_present.ValueOrDefault() && !x().Ok()) return false;
```
EricRahm, line 4394: Although note we don't use it here.

testdata/golden_cpp/condition.emb.h
```diff
@@ -4265,28 +4386,48 @@ class GenericCorrectNestedConditionalView final {
...
+    }
 
-    if (!has_xc().Known()) return false;
-    if (has_xc().ValueOrDefault() && !xc().Ok()) return false;
 
+    {
+      const auto emboss_reserved_local_field_present = ::emboss::support::Equal</**/::std::int32_t, bool, ::std::int32_t, ::std::int32_t>((emboss_reserved_local_ok_subexpr_1.Ok()    ? ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(emboss_reserved_local_ok_subexpr_1.UncheckedRead()))    : ::emboss::support::Maybe</**/::std::int32_t>()), ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(0LL)));
```
EricRahm, line 4399: ...but we do use it here

testdata/golden_cpp/condition.emb.h
```diff
@@ -4265,28 +4386,48 @@ class GenericCorrectNestedConditionalView final {
...
+      if (!emboss_reserved_local_field_present.Known()) return false;
+      if (emboss_reserved_local_field_present.ValueOrDefault() && !xc().Ok()) return false;
+    }
 
-    if (!has_xcc().Known()) return false;
-    if (has_xcc().ValueOrDefault() && !xcc().Ok()) return false;
 
+    {
+      const auto emboss_reserved_local_field_present = ::emboss::support::And</**/bool, bool, bool, bool>(::emboss::support::Equal</**/::std::int32_t, bool, ::std::int32_t, ::std::int32_t>((emboss_reserved_local_ok_subexpr_1.Ok()    ? ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(emboss_reserved_local_ok_subexpr_1.UncheckedRead()))    : ::emboss::support::Maybe</**/::std::int32_t>()), ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(0LL))), ::emboss::support::Equal</**/::std::int32_t, bool, ::std::int32_t, ::std::int32_t>((xc().Ok()    ? ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(xc().UncheckedRead()))    : ::emboss::support::Maybe</**/::std::int32_t>()), ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(0LL))));
```
EricRahm, line 4406: ...and here. Note how verbose this has become though. Maybe that's fine, this is generated code after all.

compiler/back_end/cpp/generated_code_templates
AaronWebster: Check out https://github.com/google/emboss/pull/235 which I think is more correct as to what Ben suggested I implement and the code is much cleaner.  Some compiler specific attributes could be elided, let me know.

AaronWebster:
Eric, can you please move your comments to https://github.com/google/emboss/pull/235 so we can have the discussion there?

PR #235, opened 2025-12-02
Optimize C++ generated code for structures with many conditional fields

AaronWebster:
This change improves the performance of the generated C++ code for Emboss structures containing many conditional fields.

Specifically, it optimizes the validation logic (e.g., the `Ok()` method) and field access patterns where fields depend on a shared discriminant or tag. This results in faster runtime validation and reduced overhead when interacting with complex structures defined with extensive conditional logic.

Performance Impact (on testdata/many_conditionals.emb with 10,000 iterations x 100 tags):

Runtime Speedup: ~8.7x faster (from ~1.83s to ~0.21s).
Binary Size Reduction: ~47% smaller (from 57848 bytes to 30472 bytes for many_conditionals_benchmark with -c opt).


compiler/back_end/cpp/run_one_golden_test.py
```diff
@@ -34,12 +34,24 @@ def main(argv):
     emboss_compiler = argv[2]
     emb_file = argv[3]
     golden_file = argv[4]
-    include_dirs = argv[5:]
+
+    include_dirs = []
+    compiler_flags = []
+    for arg in argv[5:]:
+        if arg.startswith("--import-dir="):
```
robrussell, line 41: This isn't very robust but I don't think `import-dir` actually works as intended. I just filed #236 , maybe the flag should be removed since it isn't used in current tests.

robrussell (approved):

compiler/back_end/cpp/generated_code_templates
```diff
@@ -124,15 +124,16 @@ class Generic${name}View final {
 
   ${enum_usings}
 
-  bool Ok() const {
+  [[nodiscard]] bool Ok() const {
```
EricRahm, line 127: I appreciate adding these annotations, but maybe we should pull them out into a separate issue and PR to discuss them in isolation. I believe it's a C++ 17 feature and we may need to hash out some details around supporting (or not) C++14.

compiler/back_end/cpp/header_generator.py
```diff
@@ -768,15 +768,24 @@ def render_field(self, expression, ir, subexpressions):
...
     def add(self, subexpr):
+        # Check if expression exists in current or parent scopes
+        scope = self
+        while scope:
+            if subexpr in scope._subexpr_to_name:
+                return scope._subexpr_to_name[subexpr]
+            scope = scope._parent
+
+        # Not found, add to current scope
         if subexpr not in self._subexpr_to_name:
```
EricRahm, line 789: nit: I don't think we need this check anymore (it's covered above).

compiler/back_end/cpp/header_generator.py
```diff
@@ -871,7 +888,9 @@ def _render_expression(expression, ir, field_reader=None, subexpressions=None):
 
 
 def _render_existence_test(field, ir, subexpressions=None):
-    return _render_expression(field.existence_condition, ir, subexpressions)
+    return _render_expression(
+        field.existence_condition, ir, subexpressions=subexpressions
```
EricRahm, line 892: Is this change actually doing anything?

compiler/back_end/cpp/header_generator.py
```diff
@@ -1330,6 +1349,143 @@ def _cpp_field_name(name):
...
+                    "expr": field.existence_condition,
+                    "fields": [],
+                }
+                ordered_keys.append(key)
+            groups[key]["fields"].append(field)
+
+    lines = []
+    for key in ordered_keys:
+        group = groups[key]
+        if group["type"] == "switch":
```
EricRahm, line 1429: In general we maybe want to use templates for the switch and if blocks, this is a little hard to follow.

compiler/back_end/cpp/header_generator.py
```diff
@@ -1330,6 +1349,143 @@ def _cpp_field_name(name):
...
+
+    if res0.is_constant and not res1.is_constant:
+        return arg1, arg0
+    if res1.is_constant and not res0.is_constant:
+        return arg0, arg1
+
+    return None, None
+
+
+def _generate_optimized_ok_method_body(fields, ir, subexpressions):
```
EricRahm, line 1395: This should probably have a big comment explaining what it's intending to do. If I understand correctly it's trying to take something like this emboss def:
```
struct blah:
  0 [+4] UInt tag
  if tag == 0:
    4 [+4] UInt var_0
  if tag == 1:
    4 [+4] UInt var_1
```

and emit
```c++
{
  const auto emboss_reserved_switch_discrim = ...;
  if (!emboss_resevered_switch_discrim.Known()) return false;
  switch (emboss_reserved_switch_discrim.ValueOrDefault()) {
    case 0: if (!var_0.Ok()) return false; break;
    case 1: if (!var_1.Ok()) return false; break;
  }
```

I might be misunderstanding what we're doing (hence having a big ol comment would be helpful).

Does this work for something like:
```
struct blah:
  0 [+4] UInt tag
  if tag == 0:
    4 [+4] UInt var_0
  if tag == 1:
    4 [+4] UInt var_1
  if tag == 0:
    8 [+4] UInt var_x
```

testdata/golden_cpp/condition.emb.h
```diff
@@ -357,48 +357,50 @@ class GenericBasicConditionalView final {
...
-
-
+      const auto emboss_reserved_local_ok_subexpr_1 = x();
+      const auto emboss_reserved_local_ok_subexpr_2 = (emboss_reserved_local_ok_subexpr_1.Ok()    ? ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(emboss_reserved_local_ok_subexpr_1.UncheckedRead()))    : ::emboss::support::Maybe</**/::std::int32_t>());
+
+    {
+      const auto emboss_reserved_cond = ::emboss::support::Maybe</**/bool>(true);
+      if (!emboss_reserved_cond.Known()) return false;
+      if (emboss_reserved_cond.ValueOrDefault()) {
+        if (!x().Ok()) return false;
```
EricRahm, line 370: So we're no longer doing `Known` and `ValueOrDefault` checks here, I'm not sure if that's intended or not. Is the idea that `has_x()`, `has_IntrinsicSizeInBytes()`, etc all boil down to `::emboss::support::Maybe</**/bool>(true);` so we can lump them together?

EricRahm (commented):
Just some initial thoughts, what you're doing seems pretty cool! It'd just help to document the intent a little more.

PR #237, opened 2025-12-03
Add JSON output support to Emboss text format

AaronWebster:
Add JSON output support to Emboss text format


compiler/back_end/cpp/generated_code_templates
```diff
@@ -1,17 +1,3 @@
-// Copyright 2019 Google LLC
```
reventlov, line 1: Unless Google has change its policies in the last few months, this header should stay.

compiler/back_end/cpp/generated_code_templates
```diff
@@ -29,14 +15,14 @@
  */
 #ifndef ${header_guard}
 #define ${header_guard}
-#include <stdint.h>
```
reventlov, line 32: Either delete if these are truly unnecessary, or uncomment if they're used in here.

compiler/back_end/cpp/generated_code_templates
```diff
@@ -29,14 +15,14 @@
...
 
-#include <algorithm>
-#include <type_traits>
-#include <utility>
+// #include <algorithm>
+// #include <type_traits>
+// #include <utility>
 
-#include "runtime/cpp/emboss_cpp_util.h"
+#include "third_party/emboss/runtime/cpp/emboss_cpp_util.h"
```
reventlov: This is an internal-ism.  Copybara will adjust the path at import.

compiler/back_end/cpp/generated_code_templates
```diff
@@ -323,20 +322,31 @@ ${write_fields}
...
+        }
         if (emboss_reserved_local_field_options.multiline()) {
           emboss_reserved_local_stream->Write(
               emboss_reserved_local_field_options.current_indent());
         } else {
-          if (emboss_reserved_local_wrote_field) {
-            emboss_reserved_local_stream->Write(",");
+          if (!emboss_reserved_local_field_options.json() ||
+              emboss_reserved_local_wrote_field) {
+            emboss_reserved_local_stream->Write(" ");
```
reventlov: You have moved the prefix `Write(" ")` so that it no longer occurs for the first element in a structure.  I don't see how the `SkippedStructureFieldOutput` is still passing.

Also: the `Write(" ")` here does *not* happen for JSON, but there are spaces between fields in the JSON goldens in the new tests.

Did I miss something, or is there version skew between the test and implementation?

compiler/back_end/cpp/testcode/text_format_test.cc
```diff
@@ -93,6 +93,89 @@ TEST(TextFormat, UpdateFromText) {
...
+TEST(TextFormat, JsonOutput) {
+  ::std::array<char, 57> values = {};
+  ::std::iota(values.begin(), values.end(), 0);
+
+  const auto view = MakeJsonTestStructView(&values);
+  EXPECT_EQ(
+      "{\"one_byte_enum\": \"ZERO\", \"seven_bit_uint\": 1, \"one_bit_flag\": "
+      "false, \"one_byte_uint\": 2, \"two_byte_uint\": 1027, "
+      "\"four_byte_uint\": 134678021, \"eight_byte_uint\": "
+      "1157159078456920585, \"uint8_array\": [17, 18, 19, 20, 21, 22, 23, 24, "
```
reventlov: So: a very "fun" thing about JSON is that all numeric values are 64-bit *floating-point* values.  The `"eight_byte_uint"` value here will not be parsed correctly by many JSON parsers, depending how closely they hew to the JSON specification.

The typical workaround for this is to serialize integers as base 10 ASCII strings, either:

*   All integers are serialized as strings (this is what Emboss does internally for the serialized IR)
*   Integers whose range is larger than +/-2^53 are serialized as strings (so any 8-byte integer would be serialized as a string, no matter what its current value)
*   Integers whose actual value is outside of [-2^53,+2^53] are serialized as strings (so an 8-byte integer might be serialized as a string or a double, depending on its current value)

I don't think there is a single correct answer here, so it might just need another config value.

compiler/back_end/cpp/generated_code_templates
```diff
@@ -295,11 +287,18 @@ ${write_fields}
     // Avoid unused variable warnings for empty structures:
     (void)emboss_reserved_local_wrote_field;
     if (emboss_reserved_local_options.multiline()) {
+      if (emboss_reserved_local_wrote_field &&
+          emboss_reserved_local_options.json()) {
```
reventlov, line 299: You can delete the `&& ...json()` here and then lines 348-351 entirely, I believe.

compiler/back_end/cpp/generated_code_templates
```diff
@@ -323,20 +322,31 @@ ${write_fields}
...
-            emboss_reserved_local_stream->Write(",");
+          if (!emboss_reserved_local_field_options.json() ||
+              emboss_reserved_local_wrote_field) {
+            emboss_reserved_local_stream->Write(" ");
           }
-          emboss_reserved_local_stream->Write(" ");
         }
-        emboss_reserved_local_stream->Write("${field_name}: ");
+        if (emboss_reserved_local_field_options.json()) {
+          emboss_reserved_local_stream->Write("\"${field_name}\": ");
```
reventlov: Most JSON serializers will omit the space between `:` and the field value when generating compact output.

runtime/cpp/emboss_text_util.h
```diff
@@ -80,12 +83,19 @@ class TextOutputOptions final {
...
+    result.json_ = new_value;
+    return result;
+  }
+
   ::std::string current_indent() const { return current_indent_; }
   ::std::string indent() const { return indent_; }
   bool multiline() const { return multiline_; }
   bool digit_grouping() const { return digit_grouping_; }
   bool comments() const { return comments_; }
   ::std::uint8_t numeric_base() const { return numeric_base_; }
```
reventlov: These three should probably return `... && !json_` instead of having `if (comments() && !json())` and similar logic scattered around the code.

compiler/back_end/cpp/generated_code_templates
AaronWebster: That would break non-JSON multiline output, suggestion?

PR #239, opened 2025-12-03
Add Optional 128-bit Integer Support to C++ Runtime

AaronWebster:
Adds optional support for 128-bit `UInt` and `Int` fields (65–128 bits) to the C++ backend and runtime, using the compiler's native `__int128_t` / `__uint128_t` where available.

## What this does
- `prelude.emb`: `UInt` and `Int` now permit static sizes up to 128 bits (previously 64).
- The runtime detects 128-bit support via `EMBOSS_HAS_INT128` (derived from `__SIZEOF_INT128__`). When it is unavailable, fields wider than 64 bits fail to compile, exactly as before.
- The concrete 128-bit type is supplied through the `EMBOSS_INT128_T` / `EMBOSS_UINT128_T` macros, which a client may override (e.g. to a library 128-bit type) before including the runtime.
- `LeastWidthInteger`, `BitBlock`, `MemoryAccessor`, and `ByteSwap` are extended to 128 bits. Reads and writes work for full-width, non-full-width (e.g. 72/96-bit), little- and big-endian, arrays, and the text format.

## Scope / limitations
- Anonymous `bits` types remain limited to 64 bits (widening them would require further work in `OffsetBitBlock` and the constraints checker).
- 128-bit support requires a platform/compiler with `__int128` (64-bit GCC/Clang). 32-bit targets and MSVC are unaffected and continue to reject >64-bit fields at compile time.

## Testing
- Runtime unit tests for `ByteSwap`, `MaskToNBits`, `IsPowerOfTwo`, `LeastWidthInteger`, `MemoryAccessor` (LE/BE, non-full-width), and `UIntView` / `IntView` read/write/sign-extension at 128 bits.
- Generated-code integration test (`int128_sizes.emb`) covering little-/big-endian structs, arrays, min/max, `CopyFrom`, out-of-range rejection, and text round-trips.

## Changes since the previous review
This revision rebases onto current `master` and fixes correctness issues in the write/validation path that the earlier version did not handle:

- **Range-check truncation.** `UIntView` / `IntView::CouldWriteValue` compared the candidate value against the field bounds after casting it to `uint64_t` / `int64_t`. With a 128-bit `ValueType` this truncated the value, so an out-of-range value for a 65–127-bit field was accepted, and `TryToWrite` then aborted (or silently truncated) instead of returning `false`. The comparison is now done in the widest supported integer type.
- **`is_signed` vs `__int128`.** `std::is_signed<__int128_t>` is `false` under a strict `-std=c++NN`, so the signed lower-bound check (and negative-value text parsing) was compiled out for 128-bit signed fields. Signedness is now determined via `std::numeric_limits<>::is_signed`, consistent with how these methods are enabled.
- Reverted two unrelated whitespace-only changes flagged in the earlier review, defined the 128-bit `ByteSwap` in terms of `EMBOSS_UINT128_T` for override consistency, and sorted the `testdata/BUILD` entry.


runtime/cpp/emboss_defines.h
```diff
@@ -311,4 +311,14 @@
 #define EMBOSS_SYSTEM_IS_TWOS_COMPLEMENT 0
 #endif  // !defined(EMBOSS_SYSTEM_IS_TWOS_COMPLEMENT)
 
+// Detect 128-bit integer support.
+// __uint128_t and __int128_t are available on GCC and Clang on 64-bit targets.
+#if !defined(EMBOSS_HAS_INT128)
```
jasongraffius, line 316: For extensibility, we could define `EMBOSS_INT128_T` and `EMBOSS_UINT128_T` type macros defined to be `__int128_t` and `__uint128_t` if not already defined and we have `__SIZEOF_INT128__`. That way client code can supply or override these to enable 128-bit integer support with a custom type or a library type not called `__int128_t` (for example, `absl::int128`).

runtime/cpp/emboss_defines.h
```diff
@@ -245,7 +245,7 @@
 // GCC's attribute syntax disallows parentheses in that particular position.
 #if !defined(EMBOSS_ALIAS_SAFE_POINTER_CAST)
 #define EMBOSS_ALIAS_SAFE_POINTER_CAST(t, x) \
-  reinterpret_cast<t __attribute__((__may_alias__)) *>((x))
+  reinterpret_cast<t __attribute__((__may_alias__))*>((x))
```
jasongraffius, line 248: This seems unrelated? Should probably revert to not pollute the diff.

runtime/cpp/emboss_cpp_types.h
```diff
@@ -21,6 +21,8 @@
 #include <cstdint>
 #include <type_traits>
 
+#include "runtime/cpp/emboss_defines.h"
```
jasongraffius, line 24: I think the relative import here should just be "emboss_defines.h" since it's in the same directory.

compiler/back_end/cpp/testcode/int128_sizes_test.cc
```diff
@@ -0,0 +1,342 @@
+// Copyright 2024 Google LLC
```
jasongraffius, line 1: 2024 -> 2026

testdata/__init__.py
```diff
@@ -1,13 +0,0 @@
-# Copyright 2019 Google LLC
```
jasongraffius, line 1: I don't think we want to delete this copyright header

testdata/int128_sizes.emb
```diff
@@ -0,0 +1,85 @@
+# Copyright 2024 Google LLC
```
jasongraffius, line 1: 2024 -> 2026

AaronWebster:
@jasongraffius Ready for another look

compiler/front_end/constraints_test.py
```diff
@@ -235,9 +235,10 @@ def test_struct_field_too_big_for_type(self):
         )
 
     def test_bits_field_too_big_for_type(self):
+        # UInt now supports up to 128 bits, so test with 129 bits (17 bytes)
```
robrussell: nit: "now" won't be relevant after this PR is in - so "UInt supports...", same below.

testdata/int128_sizes.emb
```diff
@@ -0,0 +1,85 @@
...
+  18  [+20]  UInt:80[2]   ten_byte
+  38  [+22]  UInt:88[2]   eleven_byte
+  60  [+24]  UInt:96[2]   twelve_byte
+  84  [+26]  UInt:104[2]  thirteen_byte
+  110 [+28]  UInt:112[2]  fourteen_byte
+  138 [+30]  UInt:120[2]  fifteen_byte
+  168 [+32]  UInt:128[2]  sixteen_byte
+
+
+# Note: Anonymous `bits` types are still constrained to 64 bits maximum.
```
robrussell: This comment would be more useful in the docs, maybe in the roadmap.

compiler/back_end/cpp/testcode/int128_sizes_test.cc
```diff
@@ -0,0 +1,426 @@
...
+TEST(Int128SizesView, TextFormatRoundTripNegative) {
+  ::std::uint8_t buffer[100] = {};
+  auto writer = Int128SizesWriter(buffer, sizeof buffer);
+
+  // Negative value into a 72-bit signed field: the '-' must be honored even
+  // though ::std::is_signed<__int128_t> is false under a strict -std.
+  EXPECT_TRUE(::emboss::UpdateFromText(writer, "{nine_byte: -12345}"));
+  EXPECT_EQ(static_cast<__int128_t>(-12345), writer.nine_byte().Read());
+
+  // A large negative 128-bit value (-2**120) does not fit in 64 bits; it must
```
robrussell: (-(2**120))

compiler/back_end/cpp/testcode/int128_sizes_test.cc
```diff
@@ -0,0 +1,426 @@
...
+  // round-trip through the decimal text format.
+  EXPECT_TRUE(::emboss::UpdateFromText(
+      writer, "{sixteen_byte: -1329227995784915872903807060280344576}"));
```
robrussell: If this value were in hex it would be easier to relate back to the comment above.

compiler/front_end/prelude.emb
```diff
@@ -25,14 +25,24 @@
 
 external UInt:
   -- UInt is an automatically-sized unsigned integer.
-  [static_requirements: $is_statically_sized && 1 <= $static_size_in_bits <= 64]
+  --
+  -- Note: Sizes above 64 bits (up to 128 bits) require platform support for
+  -- 128-bit integers (__uint128_t). The C++ backend will use EMBOSS_HAS_INT128
+  -- to detect availability. On platforms without 128-bit integer support,
+  -- fields larger than 64 bits will fail to compile.
+  [static_requirements: $is_statically_sized && 1 <= $static_size_in_bits <= 128]
```
robrussell, line 33: It looks like there's a place where a check for this this could be included at https://github.com/google/emboss/blob/b9f89e3572f815eb141782f102fcfa0bba6a034c/compiler/front_end/constraints.py#L704 but I'm not sure constraints.py would have enough information to know whether the platform supports 128-bit integers.

runtime/cpp/emboss_cpp_types.h
```diff
@@ -51,12 +53,26 @@ struct FloatType<32> final {
...
+template <>
+struct LeastWidthInteger<128> final {
+  using Unsigned = EMBOSS_UINT128_T;
+  using Signed = EMBOSS_INT128_T;
+};
+#else
 template <int kBits>
 struct LeastWidthInteger final {
   static_assert(kBits <= 64, "Only bit sizes up to 64 are supported.");
   using Unsigned = typename LeastWidthInteger<kBits + 1>::Unsigned;
```
robrussell, line 82: I'm a little confused about this (including the previously existing implementation)
If `kBits` is 64 then this is 

```
using Unsigned = typename LeastWidthInteger<65>::Unsigned;
```

Is this `using` just never applicable because the specializations for 64, 32, 16, and 8 always apply? Or is there something else I'm missing?

robrussell (changes requested):
The docs at https://github.com/google/emboss/blob/master/doc/language-reference.md#uint still say max size of `Int` and `UInt` are both 64 bits.

The different limit for the size of anonymous bit fields should be mentioned at https://github.com/google/emboss/blob/master/doc/language-reference.md#anonymous-bits .

PR #241, opened 2026-01-29
Optimize C++ generated code for structures with many conditional fields

AaronWebster:
This change improves the performance of the generated C++ code for Emboss structures containing many conditional fields.

Specifically, it optimizes the validation logic (e.g., the `Ok()` method) and field access patterns where fields depend on a shared discriminant or tag. This results in faster runtime validation and reduced overhead when interacting with complex structures defined with extensive conditional logic.

## Performance Impact

Benchmarked on `testdata/many_conditionals.emb` with 10,000 iterations x 100 tags, compiled with `-c opt`:

| Metric | Master | PR | Improvement |
|--------|--------|-----|-------------|
| **Compiled Binary Text Section** | 64,519 bytes | 30,079 bytes | **53.4% reduction** |
| **Runtime** (avg of 3 runs) | ~1.31s | ~0.11s | **11.8x faster** |
| **Header File Size** | 611,751 bytes | 612,180 bytes | +429 bytes (+0.07%) |

### Summary

- **Runtime Speedup: 11.8x** (from ~1.31s to ~0.11s)
- **Binary Size Reduction: 53.4%** (text section reduced from 64.5KB to 30KB, saving ~34KB)
- Header source size is essentially unchanged (trivial 0.07% increase)

---
*Recreated from #235 after the source branch was deleted.*


compiler/back_end/cpp/header_generator.py
```diff
@@ -1330,6 +1344,225 @@ def _cpp_field_name(name):
...
+          case 1:
+            if (!var_1().Ok()) return false;
+            break;
+        }
+
+    Note: When multiple fields share the same case value (like var_0 and var_x
+    both guarded by tag == 0), only the first field for that case value is
+    checked in the switch. This is because once we've checked var_0 for case 0,
+    we break out of the switch. Fields with duplicate case values will fall
+    through to the if-statement path.
```
EricRahm, line 1431: Overall great comment, I really appreciate the clear explanations. Sorry to harp on this, but can you add a test case to make sure the statements here match up with the code behavior?

I suspect this will lead to a bit more churn so I'll hold off on digging to much deeper.

compiler/back_end/cpp/one_golden_test.py
```diff
@@ -27,13 +27,15 @@ def __init__(
         emb_file,
         golden_file,
         include_dirs=None,
+        compiler_flags=None,
```
EricRahm, line 30: The one_golden_test and run_one_golden_test changes seem like they might be a separate issue? If so would you mind splitting that out (and we can land it quickly).

compiler/util/parser_types.py
```diff
@@ -15,7 +15,7 @@
 """Types related to the LR(1) parser.
 
 This module contains types used by the LR(1) parser, which are also used in
-other parts of the compiler: 
+other parts of the compiler:
```
EricRahm, line 18: nit: this seems unrelated

compiler/back_end/cpp/header_generator.py
```diff
@@ -1330,6 +1344,225 @@ def _cpp_field_name(name):
...
+            # Check if condition is statically true (constant boolean true)
+            # In this case, we can skip the if-block wrapper entirely.
+            is_always_true = (
+                cond_result.is_constant
+                and group["expr"].type.which_type == "boolean"
+                and group["expr"].type.boolean.has_field("value")
+                and group["expr"].type.boolean.value
+            )
+
+            if is_always_true:
```
EricRahm, line 1529: I love this change! I wonder if we could split it out out as a separate issue, I think it'd make the golden churn a little bit easier to grok and also highlight the impact of it :)

testdata/golden_cpp/complex_structure.emb.h
```diff
@@ -1791,84 +1763,135 @@ class GenericComplexView final {
...
+      const auto emboss_reserved_local_ok_subexpr_22 = ::emboss::support::GreaterThan</**/::std::int32_t, bool, ::std::int32_t, ::std::int32_t>(emboss_reserved_local_ok_subexpr_21, ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(0LL)));
+      const auto emboss_reserved_local_ok_subexpr_23 = ::emboss::support::And</**/bool, bool, bool, bool>(emboss_reserved_local_ok_subexpr_19, emboss_reserved_local_ok_subexpr_22);
+      const auto emboss_reserved_local_ok_subexpr_24 = ::emboss::support::LessThan</**/::std::int32_t, bool, ::std::int32_t, ::std::int32_t>(emboss_reserved_local_ok_subexpr_10, ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(2LL)));
+      const auto emboss_reserved_local_ok_subexpr_25 = ::emboss::support::LessThan</**/::std::int32_t, bool, ::std::int32_t, ::std::int32_t>(emboss_reserved_local_ok_subexpr_3, ::emboss::support::Maybe</**/::std::int32_t>(static_cast</**/::std::int32_t>(4LL)));
+      const auto emboss_reserved_local_ok_subexpr_26 = ::emboss::support::Or</**/bool, bool, bool, bool>(emboss_reserved_local_ok_subexpr_24, emboss_reserved_local_ok_subexpr_25);
```
EricRahm, line 1791: This feels like somehow this got worse, maybe the tradeoff is worth it but I wonder if we could do better.

testdata/golden_cpp/no_enum_traits.emb.h
```diff
@@ -38,79 +34,6 @@ enum class Foo : ::std::uint64_t {
   VALUE = static_cast</**/::std::int32_t>(10LL),
 
 };
-template <class Enum>
```
EricRahm, line 41: I suspect this is a result of the fix for passing in compiler args in the one_golden_test changes?

testdata/__init__.py
```diff
@@ -1,13 +0,0 @@
-# Copyright 2019 Google LLC
```
EricRahm, line 1: I don't think we want to delete this?

EricRahm (commented):
Thanks for polishing this up! Looks really good, added some high level comments. I think overall it'd be a little easier to review if we split out the always true optimization.

PR #245, opened 2026-03-04
Noncontiguous View Support

jasongraffius:
Add support for views over non-contiguous containers, allowing direct access and mutation via iterators. Iterator access is slower than contiguous containers, as currently there is no support for communicating contiguity metadata to the Emboss runtime support, but if that information is surfaced in the future as an optimization then it can defer to contiguous `MemoryAccessor`s for copies that are not across a contiguity boundary, which should improve performance.

This should not have an effect on existing contiguous containers as they continue to use contiguous memory access and `std::copy_n` will optimize to `memcpy` in all major compilers.

Contains the following commits:

---
### Introduce noncontiguous container utility

Introduce a test utility type that is a non-contiguous container that is
highly configurable in structure so we can test various conditions in
upcoming features.

---
### Add UnalignedAccessor to read/write through iterator

This enables accessing values in generic functionality that works
through any iterator, allowing transparent `memcpy` performance for
pointers (all major compilers will compile `std::copy_n` for a
`char`-like pointer into a memcpy), while supporting arbitrary
iterators.

Added tests that exhaustively test various possible non-contiguous
arrangements.

---
### Introduce IteratorStorage for any iterable

Create a storage type for arbitrary iteratable types, technically
supporting contiguous buffers, though continue to keep contiguous
buffers as it is much more performant for contiguous access.

---
### Add support for IteratorStorage to Make[Struct]View

Enable creating a struct view over non-contiguous containers using the
Make[Struct]View generated functions.

### DIFF
--- a/compiler/back_end/cpp/testcode/iterator_view_integration_test.cc
+++ b/compiler/back_end/cpp/testcode/iterator_view_integration_test.cc
@@ -1,0 +1,225 @@
+#include <gtest/gtest.h>
+
+#include <cstddef>
+#include <vector>
+
+#include "runtime/cpp/test/util/noncontiguous_container.h"
+#include "testdata/alignments.emb.h"
+#include "testdata/condition.emb.h"
+#include "testdata/dynamic_size.emb.h"
+#include "testdata/uint_sizes.emb.h"
+#include "testdata/virtual_field.emb.h"
+
+namespace emboss {
+namespace test {
+namespace {
+
+template <typename CharT>
+void TestMakeAlignmentsViewIntegration() {
+  using ByteT = typename std::remove_const<CharT>::type;
+
+  std::vector<std::vector<ByteT>> chunks = {
+      {ByteT(0x11), ByteT(0x22)},
+      {ByteT(0x33), ByteT(0x44), ByteT(0x55)},
+      {ByteT(0x66), ByteT(0x77), ByteT(0x88)},
+      {ByteT(0x00), ByteT(0x00), ByteT(0x00), ByteT(0x00)},
+      {ByteT(0x00), ByteT(0x00), ByteT(0x00), ByteT(0x00)},
+      {ByteT(0x00), ByteT(0x00), ByteT(0x00), ByteT(0x00)},
+      {ByteT(0x00), ByteT(0x00), ByteT(0x00), ByteT(0x00)},
+      {ByteT(0x00), ByteT(0x00), ByteT(0x00), ByteT(0x00)}};
+
+  ::emboss::support::test::NoncontiguousContainer<ByteT> container(chunks);
+
+  auto view = MakeAlignmentsView(container);
+  EXPECT_TRUE(view.Ok());
+  EXPECT_EQ(view.SizeInBytes(), 28U);
+
+  EXPECT_EQ(view.zero_offset().dummy().Read(), 0x11223344U);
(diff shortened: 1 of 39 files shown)


EricRahm:
Thanks, I think this is a great addition!

Following up on an out-of-band conversation:
1. The overall diff is quite large, but a lot of that is golden updates
2. We discussed concrete use cases, one we had in mind is supporting things like [pw::multibuf](https://pigweed.dev/pw_multibuf/). @jasongraffius is going to see how well the interface we have here would work with that.
3. I had some higher level questions around C++14 support, binary size, compile time, impact on existing usage, etc. Sounds like C++14 is probably fine, Jason is going to do some adhoc tests against `pw_bluetooth` for the rest.

I'll do a cursory review, but hold off on a final push until @jasongraffius has time to confirm things still work well for `pw_bluetooth`.

runtime/cpp/test/util/BUILD
```diff
@@ -0,0 +1,18 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
```
EricRahm, line 1: nit: copyright header?

runtime/cpp/emboss_memory_util.h
```diff
@@ -80,32 +78,65 @@ struct MemoryAccessor<CharT, 1, 0, kBits> {
...
+  template <typename Iterator>
+  static inline Unsigned ReadLittleEndianUInt(Iterator bytes) {
+    using DereferenceT = typename ::std::remove_cv<
+        typename ::std::remove_reference<decltype(*bytes)>::type>::type;
+    static_assert(
+        ::std::is_same<DereferenceT,
+                       typename ::std::remove_cv<CharT>::type>::value,
+        "UnalignedAccessor Iterator underlying type must match CharT.");
     Unsigned result = 0;
-    ::std::memcpy(&result, bytes, kBits / 8);
```
EricRahm, line 85: The overall concern here would be that we're removing the use of `memcpy` in favor of `copy_n` in a few places. This might be a regression on some systems, I wonder if it's worth doing some template shenanigans to still use memcpy if we have a pointer type for the iterator. See bolms@'s note above about this code being perf sensitive.

runtime/cpp/emboss_memory_util.h
```diff
@@ -800,6 +800,205 @@ class ContiguousBuffer final {
...
+ public:
+  using ByteType = typename ::std::remove_reference<
+      decltype(*::std::declval<Iterator>())>::type;
+  using CharT = typename ::std::remove_cv<ByteType>::type;
+
+  template <::std::size_t kSubAlignment = 1, ::std::size_t kSubOffset = 0>
+  using OffsetStorageType = IteratorStorage<Iterator>;
+
+  static_assert(IsAliasSafe<CharT>::value,
+                "IteratorStorage requires a char type iterator.");
```
EricRahm, line 821: nit: "alias safe iterator"

compiler/back_end/cpp/testcode/iterator_view_integration_test.cc
```diff
@@ -0,0 +1,225 @@
+#include <gtest/gtest.h>
```
EricRahm, line 1: nit: copyright?

EricRahm (commented):
Overall looks pretty good, thanks for breaking it up into logical changes. Some pretty minor comments, my main concern is going from `memcpy` -> `copy_n`

PR #251, opened 2026-05-20
Add scripts/regenerate_goldens.py for batch golden regeneration

AaronWebster:
Mirrors the `cpp_golden_test` list in `compiler/back_end/cpp/BUILD`, invoking embossc on each input `.emb` file to regenerate the corresponding golden header. Makes generator changes (which usually require regenerating dozens of golden files) a one-command operation instead of a per-file dance.

Also handles the `testdata/imported_genfiles.emb` genrule transform locally and cleans up afterward so the produced file doesn't collide with the bazel genrule on subsequent builds.

No generator changes; no measurable impact on generated code.

### DIFF
--- a/scripts/regenerate_goldens.py
+++ b/scripts/regenerate_goldens.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+"""Regenerates testdata/golden_cpp/*.h files.
+
+Run from the repo root. Mirrors the cpp_golden_test list in
+compiler/back_end/cpp/BUILD.
+"""
+
+import os
+import subprocess
+import sys
+
+REPO = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+# Golden tests: emb file (under testdata/) -> golden file (under testdata/golden_cpp/).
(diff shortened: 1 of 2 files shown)


scripts/regenerate_goldens.py
```diff
@@ -0,0 +1,110 @@
...
+"""
+
+import os
+import subprocess
+import sys
+
+REPO = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+# Golden tests: emb file (under testdata/) -> golden file (under testdata/golden_cpp/).
+# All cpp_golden_test entries from compiler/back_end/cpp/BUILD.
```
robrussell, line 15: Is there any way to keep the two lists in sync? Trying to dynamically generate this list from the other BUILD file would probably get messy. A comment at https://github.com/google/emboss/blob/master/compiler/back_end/cpp/BUILD#L402 would be good though.

robrussell (approved):

scripts/regenerate_goldens.py
AaronWebster: Done.

PR #252, opened 2026-05-20
Add embedded text-size measurement scripts

AaronWebster:
`scripts/embedded_bench.sh` cross-compiles a tiny driver TU that pulls `LargeConditionalsView::Ok()` and `DisjunctionConditionalsView::Ok()` into an object file, then reports `size` totals and per-symbol sizes on:

  * ARM Cortex-M4 Thumb-2 (`arm-none-eabi-g++ -mthumb -mcpu=cortex-m4 -Os`)
  * MicroBlaze big-endian (`microblaze-buildroot-linux-gnu-g++ -Os`)
  * Host x86-64 `-Os` as a reference

All builds use `-ffunction-sections` + `-fdata-sections` so each function gets its own section, making `nm --size-sort` and the object's total `.text` directly comparable across compiles.

`scripts/embedded_compare.sh` runs the bench on the current branch and a base ref (default `master`), forwarding the test schema and benchmark TU through the checkout so the comparison measures the generator output and not the test surface area. Prints a side-by-side table with absolute deltas and percentages.

Requires `arm-none-eabi-g++` and a `microblaze-*` `g++` on PATH or in `/opt/microblaze`. Builds on the regen script from #251.

No generator changes; no measurable impact on generated code.

### DIFF
--- a/scripts/embedded_bench.sh
+++ b/scripts/embedded_bench.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+# Cross-compile the many_conditionals.emb.h header on embedded toolchains and
+# report .text and per-function sizes for the generated Ok() methods.
+#
+# Usage:  scripts/embedded_bench.sh <out-dir>
+# Run from the repo root after regenerating goldens / building the header.
+#
+# Each target compiles a tiny TU that forces the relevant Ok() methods to be
+# emitted, then reports `size` on the resulting object file plus a few key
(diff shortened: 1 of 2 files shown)


AaronWebster:
Superseded by #270, which folds this tooling into a single CI-integrated code-size + execution-speed harness. Its dual-Ok()-symbol coverage (LargeConditionals + DisjunctionConditionals) is preserved there.

PR #256, opened 2026-05-20
Extend switch matcher: disjunction, conjunction, single-entry demotion

AaronWebster:
PR #241's `_get_switch_candidate` only matched bare `discrim == K` equalities. Real protocol grammars express the same intent in many forms; this PR rewrites the matcher to recognize them all.

`_extract_switch_arms` decomposes an `existence_condition` into a `(discriminant, [(case_value, residual), …])` tuple, handling:

  * Bare equality (`tag == K`) — one arm, no residual.
  * Disjunction of equalities on a shared discriminant (`tag == A || tag == B || tag == C`) — one arm per `Ki`, no residual. Common for tagged unions where several values share a payload. Combined with #254's identical-body coalescing, this collapses to a single multi-label switch arm.
  * Conjunction with an equality (`tag == K && other_predicate`) — one arm carrying `[other_predicate]` as residual. Useful for nested guards like `if outer_flag && tag == K`.
  * Mixed shapes inside a disjunction: `(tag == 0 && a) || (tag == 1 && b) || tag == 2` — three arms with respective residuals `[a]`, `[b]`, `[]`.

An arm with a residual emits the `has_${field}()`-based check as its case body (the residual is folded into the existing accessor) — sound and lets the C++ compiler fold the case-pinned discriminant comparison via inlining.

A demotion pass measures total arm-entries per group and falls back to `ok_method_test` when the count is below 2 — without this, a lone field like `if outer && tag == K: xc` would get wrapped in a one-case switch whose overhead (temporary, Known() guard, scope braces) exceeds the dedupe. This also fixes a latent bug from #253: the scoped discriminant render mutated `subexpressions` during the grouping pass, so groups that later got demoted would have left dead `const auto = …;` definitions in the emitted `Ok()`. The scoped render now happens at emit time, only for surviving groups.

The benchmark schema gains a `DisjunctionConditionals` struct exercising three `||` chains (2, 3, 3 labels) and the benchmark TU gains a runtime test for it. Golden churn: `many_conditionals.emb.h` gains the new struct's output; `condition.emb.h` and `virtual_field.emb.h` shrink because several `BasicConditional`-style structs had a lone `xc` field that PR #241 was wrapping in a one-arm switch — demotion brings them back to the cheaper `has_xc()` check.

|---|---|---:|---:|---:|
| ARM Thumb-2 | TU .text | 18962 | 18208 | −754 (−4.0%) |
| ARM Thumb-2 | LargeConditionals::Ok() | 5382 | 4746 | −636 (−11.8%) |
| ARM Thumb-2 | DisjunctionConditionals::Ok() | 358 | 270 | **−88 (−24.6%)** |
| MicroBlaze | TU .text | 43640 | 42612 | −1028 (−2.4%) |
| MicroBlaze | LargeConditionals::Ok() | 14824 | 14104 | −720 (−4.9%) |
| MicroBlaze | DisjunctionConditionals::Ok() | 640 | 552 | −88 (−13.8%) |
| Host x86-64 | TU .text | 29166 | 28304 | −862 (−3.0%) |
| Host x86-64 | LargeConditionals::Ok() | 3948 | 3065 | −883 (−22.4%) |
| Host x86-64 | DisjunctionConditionals::Ok() | 431 | 375 | −56 (−13.0%) |

Incremental vs. #254: `DisjunctionConditionals::Ok()` drops 24.6% on Thumb-2, 13.8% on MicroBlaze, 13.0% on x86-64 — the matcher extensions account for nearly all of those.

Stacked on #254.

### DIFF
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -1403,6 +1403,116 @@ def _get_switch_candidate(expression, ir):
     return None, None
 
 
+def _flatten_function(expression, function):
+    """Flatten a binary function chain into a flat list of operands.
+
+    Both `(a OP b) OP c` and `a OP (b OP c)` flatten to `[a, b, c]`.
+    """
+    if (
+        expression.which_expression == "function"
+        and expression.function.function == function
+    ):
+        out = []
+        for arg in expression.function.args:
+            out.extend(_flatten_function(arg, function))
+        return out
+    return [expression]
+
+
+def _flatten_or(expression):
+    return _flatten_function(expression, ir_data.FunctionMapping.OR)
+
+
+def _flatten_and(expression):
+    return _flatten_function(expression, ir_data.FunctionMapping.AND)
+
+
+def _extract_switch_arms(expression, ir):
+    """Decomposes `expression` into a list of switch arms.
+
+    Returns either:
+      * (discriminant_expr, [(case_value_expr, residual_conjuncts), ...])
+        when the expression decomposes into a switch on a single
+        discriminant. `residual_conjuncts` is an empty list when the arm
+        is a bare equality on the discriminant; otherwise it is the list
+        of additional IR sub-expressions (`AND`-conjuncts) that the arm
+        must also satisfy. Case values are returned in encounter order;
+        the caller is responsible for sorting/coalescing.
+      * (None, None) when the expression doesn't decompose into a switch.
+
+    Supported shapes:
+      * `discrim == K`                         → one arm, no residual.
+      * `discrim == K1 || discrim == K2 || …`  → one arm per Ki, no
+        residual. Common for tagged unions where several tag values
+        share a payload (`if tag == 0x10 || tag == 0x11: …`).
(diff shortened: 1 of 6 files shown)


compiler/back_end/cpp/header_generator.py
```diff
@@ -1469,31 +1579,51 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
...
+            if id(field) not in group["encountered_field_ids"]:
+                group["encountered_field_ids"].add(id(field))
+                group["encounter_order"].append(field)
+            for case_expr, residual in arms:
+                case_str = _render_case_label(case_expr, ir)
+                case_entry = group["cases_by_label"].setdefault(
+                    case_str, {"sort_key": _case_sort_key(case_expr), "entries": []}
+                )
+                # Avoid duplicating a field across cases that map to the same
+                # case label twice (`tag == 0 || tag == 0`); after #9 the
```
robrussell, line 1622: Better to clarify if #9 is a reference to common subexpression elimination. Github bug references don't get linked in other contexts.

robrussell (approved):

PR #258, opened 2026-05-20
Tail-form return in terminal switches

AaronWebster:
When the optimized `Ok()` switch is the function's tail — last group emitted, no `[requires]` clause after — arms that validate exactly one field with no residual collapse from

    case K:
      if (!field().Ok()) return false;
      break;

to

    case K:
      return field().Ok();

Saves one conditional branch per qualifying arm. On Thumb-2 / MicroBlaze (where `-Os` is conservative about merging cmp+beq sequences) and on older compilers that don't see through the if/return-false/break shape, this trims a few bytes per case across large tagged-union schemas.

Per-arm decision: any arm that doesn't qualify (multi-field, residual, or the switch isn't terminal) keeps the break-form body. Mixing forms within one switch is safe — the default fallthrough lands on the function's `return true;` either way.

Golden churn: `many_conditionals.emb.h` shrinks substantially (the 99 single-field cases all collapse to tail-form); `condition.emb.h` and `parameters.emb.h` see smaller similar collapses.

|---|---|---:|---:|---:|
| ARM Thumb-2 | TU .text | 18962 | 17974 | **−988 (−5.2%)** |
| ARM Thumb-2 | LargeConditionals::Ok() | 5382 | 4930 | −452 (−8.4%) |
| ARM Thumb-2 | DisjunctionConditionals::Ok() | 358 | 262 | **−96 (−26.8%)** |
| MicroBlaze | TU .text | 43640 | 35796 | **−7844 (−18.0%)** |
| MicroBlaze | LargeConditionals::Ok() | 14824 | 8308 | **−6516 (−44.0%)** |
| MicroBlaze | DisjunctionConditionals::Ok() | 640 | 532 | −108 (−16.9%) |
| Host x86-64 | TU .text | 29166 | 28242 | −924 (−3.2%) |
| Host x86-64 | LargeConditionals::Ok() | 3948 | 3040 | −908 (−23.0%) |
| Host x86-64 | DisjunctionConditionals::Ok() | 431 | 354 | −77 (−17.9%) |

Note: `LargeConditionals::Ok()` measured by `nm` grows slightly on Thumb-2 and MicroBlaze for this PR alone (+302 / +604 bytes), but TU `.text` still shrinks. The compiler is inlining tail-form arms into callers more aggressively, so reduced overall code lives in `Ok()`'s callers rather than `Ok()` itself.

Stacked on #257.

### DIFF
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -1554,7 +1554,9 @@ def _extract_switch_arms(expression, ir):
     return None, None
 
 
-def _generate_optimized_ok_method_body(fields, ir, subexpressions):
+def _generate_optimized_ok_method_body(
+    fields, ir, subexpressions, allow_tail_form=False
+):
     """Generates optimized C++ code for the Ok() method body.
 
     This function optimizes validation logic for structures with conditional
@@ -1698,6 +1700,19 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
         if total_entries < 2:
             group["type"] = "demoted_to_if"
 
+    # Find the last surviving switch group; only that one is eligible for
+    # tail-form case rewrites, since tail-form turns the switch into the
+    # function's exit and any later block would be unreachable.
+    last_switch_key = None
+    if allow_tail_form:
+        for key in reversed(ordered_keys):
+            if groups[key]["type"] == "switch":
+                last_switch_key = key
+                break
+            # Any non-switch block after a switch means that switch is
+            # not the last statement in Ok(); abandon tail-form.
+            break
+
     blocks = []
     for key in ordered_keys:
         group = groups[key]
@@ -1711,6 +1726,7 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
             group["known_check_required"] = not _is_discriminant_provably_known(
                 group["discrim_expr"], fields
             )
+            group["tail_form"] = key == last_switch_key
             blocks.append(_emit_switch_block(group))
         elif group["type"] == "demoted_to_if":
             for field in group["encounter_order"]:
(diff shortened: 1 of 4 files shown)


testdata/condition.emb
```diff
@@ -229,3 +229,37 @@ struct ConditionalOnFlag:
...
+  # carries a residual (`outer == 1`), so when `outer != 1` the message is
+  # valid even though `tag` is absent -- and on a buffer too short to hold
+  # `tag`, reading the discriminant is Unknown.  The optimized Ok() must not
+  # reject such a message.
+  0 [+1]    UInt  outer
+  if outer == 1:
+    1 [+1]  UInt  tag
+
+  if tag == 0 && outer == 1:
+    2 [+1]  UInt  a
```
robrussell, line 246: nit: The counting 1,2,3 in this definition seems odd. Putting `a` and `b` in the same location (2) seems more realistic. Same below. It doesn't matter much but examples are often copy/pasted for new test cases.

testdata/condition.emb
```diff
@@ -229,3 +229,37 @@ struct ConditionalOnFlag:
...
     1 [+1]  UInt  value
+
+
+struct ResidualConditionalDiscriminant:
+  # `tag` is itself conditional, and is the switch discriminant for `a`/`b`
+  # (it is the first conjunct of their existence conditions).  Every arm
+  # carries a residual (`outer == 1`), so when `outer != 1` the message is
+  # valid even though `tag` is absent -- and on a buffer too short to hold
+  # `tag`, reading the discriminant is Unknown.  The optimized Ok() must not
+  # reject such a message.
```
robrussell, line 240: Is this a change compared to the previous `Ok()` result for a struct like this? I'm not sure what I expect below when `tag` isn't defined but the value is checked. I could be missing something important about scoping in Emboss though.

AaronWebster: The guard fix and these structs are in this PR because we propagated the fix down the stack (PR 258, 259) so it can't be reintroduced as the chain lands — that's why a behavior-affecting change shows up in what's nominally the tail-form PR.

AaronWebster:
Closing as part of the Ok()-rework restructuring. Independent re-measurement showed the tail-form return *grows* the Ok() symbol ~+7% on every target (ARM / MicroBlaze BE+LE / x86) to buy ~1% TU — a net regression on the metric this work optimizes. Its correctness precursor is now #271 (superseding #266); the grounded size wins are pursued in #272 instead.

PR #259, opened 2026-05-20
Statically-known existence conditions: drop or simplify the check

AaronWebster:
Many Emboss schemas have fields whose `existence_condition` the front end has already constant-folded to a boolean literal:

  * `if false:` fields are never present, so `Ok()` needs no check for them at all. PR #241 still emitted the full `has_X()` guard + `Ok()` check; both branches are dead.
  * Unconditional fields (and `if true:` fields) carry a literal `true` existence. `has_X()` is always `Known` and always true, so the standard `ok_method_test` pattern collapses from two lines to one bare `if (!x().Ok()) return false;` check.

A fast pre-pass in `_generate_optimized_ok_method_body` recognizes both shapes and routes them to the short emit paths. `if false:` fields are skipped; the always-true case emits the bare `Ok()` check. All unconditional fields share one group so they emit contiguously at the position of the first such field — preserves PR #241's field-batching shape so the compiler's view of the function body stays tight.

The docstring on `_generate_optimized_ok_method_body` is rewritten to describe the four emit paths the function now routes fields through (skipped / unconditional / switch arm / `ok_method_test`).

|---|---|---:|---:|---:|
| ARM Thumb-2 | TU .text | 18962 | 17974 | −988 (−5.2%) |
| ARM Thumb-2 | LargeConditionals::Ok() | 5382 | 4930 | −452 (−8.4%) |
| ARM Thumb-2 | DisjunctionConditionals::Ok() | 358 | 262 | −96 (−26.8%) |
| MicroBlaze | TU .text | 43640 | 35796 | −7844 (−18.0%) |
| MicroBlaze | LargeConditionals::Ok() | 14824 | 8308 | −6516 (−44.0%) |
| MicroBlaze | DisjunctionConditionals::Ok() | 640 | 532 | −108 (−16.9%) |
| Host x86-64 | TU .text | 29166 | 28242 | −924 (−3.2%) |
| Host x86-64 | LargeConditionals::Ok() | 3948 | 3040 | −908 (−23.0%) |
| Host x86-64 | DisjunctionConditionals::Ok() | 431 | 354 | −77 (−17.9%) |

**Incremental delta vs. #258: 0 across the board.** The C++ compiler optimizes both source patterns identically at `-Os`, so this PR is a source-quality improvement (32 golden headers shrink by hundreds of lines apiece, generated code is easier to read) with no runtime machine-code cost.

Stacked on #258.

### DIFF
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -1557,68 +1557,53 @@ def _extract_switch_arms(expression, ir):
 def _generate_optimized_ok_method_body(
     fields, ir, subexpressions, allow_tail_form=False
 ):
-    """Generates optimized C++ code for the Ok() method body.
+    """Generates the field-validation body of a structure's Ok() method.
 
-    This function optimizes validation logic for structures with conditional
-    fields by grouping fields that share the same discriminant into switch
-    statements, rather than generating separate if-statements for each field.
+    For each conditional field, routes it to one of four emit paths:
 
-    For example, given this Emboss definition:
+      1. **Skipped**. `if false:` fields can never be present, so Ok()
+         needs no check for them at all.
+      2. **Unconditional**. Non-conditional and `if true:` fields skip
+         the has_${field}() wrapper and emit a bare
+         `if (!field().Ok()) return false;` — has_X() is statically
+         Known and always true.
+      3. **Switch arm**. Fields whose existence condition decomposes
+         into a discriminant test (`discrim == K`, a disjunction of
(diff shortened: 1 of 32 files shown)


AaronWebster:
Closing as part of the Ok()-rework restructuring. Independent re-measurement showed this produces 0 machine-code change on every target (ARM / MicroBlaze BE+LE / x86) — the compiler already folds statically-known existence conditions — so it is readability-only. Dropping rather than rebasing the slice onto the new chain (#271 correctness fix, #272 residual-read elimination).

PR #265, opened 2026-06-04
Add embedded code-size benchmarking scripts

AaronWebster:
## What

Adds two developer scripts under `scripts/` for measuring the static code size of
Emboss-generated C++ across embedded toolchains and git revisions, so the ongoing
`Ok()` / switch code-generation work can be quantified on real targets.

### `scripts/embedded_bench.sh [out-dir]`

Cross-compiles a tiny freestanding TU that forces `LargeConditionals::Ok()` (from
`testdata/many_conditionals.emb`, a 100-way `tag` switch) into an object file, then
reports the object's `size` totals and the `Ok()` symbol size via `nm --size-sort`
on:

- ARM Cortex-M4 / Thumb-2 (`arm-none-eabi-g++`)
- MicroBlaze big-endian (`microblaze-buildroot-linux-gnu-g++`)
- Host x86-64 (`g++`, reference)

`-ffunction-sections`/`-fdata-sections` keep the per-symbol and TU numbers
comparable across compiles; `EMBOSS_BENCH_FLAGS` overrides the compiler flags; and
any missing toolchain is skipped with a warning rather than failing the run. The
measured header is refreshed from the live generator via
`scripts/regenerate_goldens.py`, so only the out-dir is written.

### `scripts/profile_tool.py --revisions <a> <b> ... [--configs cfg.json] [--out-dir dir]`

Drives `embedded_bench.sh` across multiple git revisions and compiler
configurations (default `Os`/`O2`/`O0`). It pulls the schema and harness forward
from the starting revision so the comparison measures generator output rather than
the test surface, then writes `<out-dir>/profile_report.md` containing, for each
target and optimization level, the TU and per-`Ok()`-symbol sizes for every
revision with deltas against the baseline revision. Requires a clean tree and
restores the original branch on exit (including on failure).

## Notes

- No production or build code changes; `scripts/` are standalone dev tools and are
  not part of any Bazel target.
- The fixtures (`testdata/many_conditionals.emb`, its golden, and
  `scripts/regenerate_goldens.py`) already exist on `master`.

### DIFF
--- a/scripts/embedded_bench.sh
+++ b/scripts/embedded_bench.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Cross-compile the generated many_conditionals.emb.h header on embedded
+# toolchains and report .text and per-function sizes for the generated Ok()
+# methods.
+#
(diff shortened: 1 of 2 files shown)


AaronWebster:
Superseded by #270: embedded_bench.sh is replaced by scripts/size_bench.py, which adds MicroBlaze LE (Vitis authoritative + Bootlin proxy) and a dynamic retired-instruction speed metric.

PR #266, opened 2026-06-04
Guard conditional switch discriminants in optimized Ok()

AaronWebster:
## Problem

The optimized `Ok()` switch (introduced in #253, extended in #256/#257) reads its
discriminant once and emits, as a safety net:

```cpp
const auto emboss_reserved_switch_discrim = <discriminant>;
if (!emboss_reserved_switch_discrim.Known()) return false;
switch (emboss_reserved_switch_discrim.ValueOrDefault()) { ... }
```

When the discriminant is itself a **conditionally-present field** and is read from a
buffer too short to contain it, the discriminant reads as `Unknown`, so this guard
makes `Ok()` return `false` — even though the message is valid, because the fields
gated on the (absent) discriminant don't exist.

This is latent on `master`: no existing test schema produces the shape, and the only
surviving guard in the goldens (`parameters.emb`'s `Axis`) is on a bare arm, which is
sound.

## Root cause

The guard is sound **iff** the group has at least one *bare* arm (existence condition
exactly `discriminant == K`): then an `Unknown` discriminant leaves that field's
`has_X()` `Unknown`, so the unoptimized `Ok()` bails for the same reason. When **every**
arm carries a residual (extra conjuncts beyond `discriminant == K`), a valid message can
have an out-of-bounds (`Unknown`) discriminant while every arm's residual is statically
false — making every `has_X()` `Known`-false (absent). The blanket guard wrongly rejects it.

## Fix

In `_emit_switch_block`, classify each group:

1. **Provably-present discriminant** → switch, no guard (unchanged, #257).
2. **Not provably-present, ≥1 bare arm** → switch + `Known()` guard (unchanged — sound).
3. **Not provably-present, all arms have residuals** → guarded switch: dispatch only when
   the discriminant is in bounds, otherwise fall back to per-field `has_X().Known()` checks.
   (An `Unknown` discriminant can never make `has_X()` `Known`-true, so the value check is
   unnecessary in the fallback.)

Cases 1 & 2 are untouched, so **every existing golden is byte-for-byte identical**; only the
new structs add codegen.

## Tests

Adds two structs to `testdata/condition.emb`:

- `ResidualConditionalDiscriminant` — the bug (all-residual arms on a conditional discriminant).
- `BareConditionalDiscriminant` — the sound-guard case (bare arms).

with regression tests in `condition_test.cc` covering: discriminant absent (valid, was
wrongly rejected), discriminant present-but-truncated (invalid), gated field present/truncated,
and discriminant matching no arm. The tests pass under **both** the optimized and unoptimized
(`_no_opts`) `Ok()` implementations, proving the optimized path now matches ground truth.

The same fix + tests will be propagated to the open stacked PRs (#258, #259) so the bug cannot
be reintroduced as they land.

### DIFF
--- a/compiler/back_end/cpp/generated_code_templates
+++ b/compiler/back_end/cpp/generated_code_templates
@@ -411,6 +411,26 @@ ${switch_cases}
     }
 
 
+// ** ok_method_switch_block_guarded ** ////////////////////////////////////////
+    // Check fields that share a conditionally-present discriminant.  Every arm
+    // carries a residual, so when the discriminant is out of bounds the message
+    // is still valid for any arm whose residual is statically false.  Dispatch
+    // only when the discriminant can be read; otherwise fall back to per-field
+    // existence checks.  (An Unknown discriminant can never make has_X()
+    // Known-true, so only the Known() test is needed in the fallback.)
+    {
+      const auto emboss_reserved_switch_discrim = ${discriminant};
+      if (emboss_reserved_switch_discrim.Known()) {
+        switch (emboss_reserved_switch_discrim.ValueOrDefault()) {
+${switch_cases}
+
+          default: break;
+        }
+      } else {
+${fallback_checks}      }
+    }
+
+
 // ** ok_method_switch_arm ** //////////////////////////////////////////////////
 ${case_labels}${case_body}          break;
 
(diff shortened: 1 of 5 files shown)


AaronWebster:
Superseded by #271, which fixes the same unsound guard more simply — demoting all-residual, not-provably-known switch groups to per-field has_X() checks instead of a guarded-switch template. #266's regression tests are carried over verbatim.

PR #268, opened 2026-07-07
Propose Rust backend design

jasongraffius:
Proposal and design for a Rust backend for Emboss

### DIFF
--- a/doc/design_docs/rust_backend.md
+++ b/doc/design_docs/rust_backend.md
@@ -1,0 +1,650 @@
+# Rust Code Generation Backend for Emboss
+
+## Motivation
+
+Emboss currently supports generating C++ code for parsing and validating binary data structures. As Rust becomes more widely used for systems programming and protocol implementations, there is a growing need for native Rust support in Emboss.
+
+This document proposes a design for an Emboss Rust backend. The goal is to generate Rust code that provides the same safety, zero-copy, and efficiency guarantees as the C++ backend, while integrating naturally with Rust's type system and safety features.
+
+## Requirements
+
+A Rust backend for Emboss should satisfy the following requirements:
+
+1.  **Zero-copy**: Accessing fields should not require copying data into intermediate structures.
+2.  **Safety**: The generated API must be memory-safe. Any use of `unsafe` must be internal to the runtime, minimized, and thoroughly verified.
+3.  **Idiomatic Rust**: The API should feel natural to Rust developers, utilizing features like `Result`, `Option`, and traits.
+4.  **`no_std` and `no_alloc` Support**: The core parsing functionality and generated code must advance embedded systems use cases by supporting `#![no_std]` and requiring no heap memory allocation. Auxiliary features (like text serialization) may optionally require `std`.
+
+## Proposed Design
+
+### Views and Storage
+
+Similar to the C++ backend, the core concept of the Rust backend is a "view" over backing storage. A view does not own the data; it references a backing storage. The main view type representing a generated structure is referred to as the **base view**.
+
+#### Storage Genericity
+The generated views are generic over the storage type `S: Storage`. While `&[u8]` is the default slice-backed storage implementation, any custom storage backend (e.g. ring buffers, page-aligned buffers, or scatter-gather lists) that implements the `Storage` trait is fully supported.
+
+#### Copy/Clone Semantics
+A view inherits the `Copy` and `Clone` capabilities of its underlying storage type:
+*   A view using a shared slice, `BasicType<&'a [u8]>`, naturally implements `Copy` and `Clone`.
+*   A writer view using a mutable slice, `BasicType<&'a mut [u8]>`, does **not** implement `Copy` or `Clone` and is move-only, satisfying Rust's exclusive access rules.
+
+#### Aliasing
+All access and borrow safety is managed transparently via standard Rust borrows and lifetimes. Multiple read-only views can borrow the same buffer concurrently, while writer views require exclusive access, enforced at compile time by Rust's borrow checker. No custom unsafe compiler trickery is introduced.
+
+#### Passing Nested Views (View Detachment)
+Helper functions that process sub-structures should be generic over the storage to accept nested views:
+```rust
+fn process_payload<S: Storage>(payload: Payload<S>) { ... }
+```
+If a storage type supports sub-slicing (producing a slice of the parent's storage), a sub-view can be "detached" from its nested wrapper (`NestedStorage<Outer>`) to yield a flat view backed directly by the parent's slice type (e.g., yielding `Payload<&[u8]>` from a `Parent<&[u8]>`). The Emboss runtime remains agnostic to how the storage slice is implemented.
+
+Consider the simple `BasicType` definition:
+
+```emboss
+[$default byte_order: "BigEndian"]
+
+struct BasicType:
+  0 [+2]  UInt  a
+  2 [+2]  UInt  b
+```
+
+A reader view can be constructed and accessed as follows:
+
+```rust
+// Example usage of a generated view
+let buf = [0x01, 0x0A, 0x02, 0x0B];
+let view = BasicType::new(&buf);
+assert_eq!(view.a().try_read(), Ok(0x010A));
+```
+
+
+### Escape Hatches
+
+When developers need to drop out of the native Emboss typed API and interact with underlying data:
+1. **Accessing Storage:** The underlying storage can be borrowed via `.storage()` on any view, or taken by value via `.into_storage()`.
+2. **Reading/Writing Beyond Boundaries:** If you need to read bytes outside the bounds of the current logical view, you can extract the storage using the methods above and perform slice operations manually. This prevents out-of-bounds access natively through the view interfaces while supporting flexible manual manipulation.
+3. **Reattaching Views:** Because the `Storage` trait is implemented for native slice types like `&[u8]`, developers can manually slice a buffer and construct a new, independent view over that specific slice whenever needed.
+
+### Accessor API
+
+In the initial implementation, the API prioritizes simplicity and safety. Fields cannot be accessed directly without runtime bounds checks. 
+
+Read and write accessors perform dynamic bounds checking on the fly and return `Result` types. A view constructed over mutable storage (`MutStorage` such as `&mut [u8]`) naturally supports modification without requiring specialized writer types.
+
+For example, using the `BasicType` defined above:
+
+```rust
+let mut buf = [0u8; 4];
+let mut view = BasicType::new(&mut buf);
+
+view.a().try_write(0x070A)?;
+view.b().try_write(0x080B)?;
+
+let val = view.a().try_read()?;
+```
+
+### Versioning and Support
+
+* Initial implementation will land clearly marked `experimental`, and should not be relied upon to not change yet. During that period we will target the 2024 edition, though we don't intend to rely on features that would preclude earlier edition support.
+* Once we stabilize the implementation, we'll declare it as such and remove the experimental marker.
+* We will make a decision on support timelines at that time, but suggest a 3 year window, essentially the current edition and previous edition (which are on 3 year tracks), as the Rust ecosystem and tooling tends to support the latest versions quickly and provide tools for easy upgrading.
+* That said, that support window is subject to change pending additional discussion and final decision is out of scope of this proposal.
+* During the experimental period, intentional breaking changes should be communicated to the emboss mailing list.
+
+### Naming Conventions
+
+Emboss's enforced naming conventions align with Rust conventions:
+*   `CamelCase` for struct/enum types maps to Rust `CamelCase` structs/enums.
+*   `snake_case` for fields maps to Rust `snake_case` methods.
+*   `SHOUTY_CASE` for enum values and constants maps to Rust `SHOUTY_CASE` constants.
+*   **Rust Keywords**: If an Emboss field or type name collides with a Rust keyword (e.g., `type`, `match`), the generator will escape it using Rust's raw identifier syntax (e.g., `r#type`).
+
+#### Trait Disambiguation
+Emboss structures can have fields with arbitrary names. To avoid naming conflicts between generated field accessors and built-in helper methods (like `size_in_bytes` or `storage`), helper methods are defined on traits (e.g., `emboss_runtime::StructView`).
+
+Disambiguation is achieved by calling the trait method directly:
+```rust
+// Accessing a field named `size_in_bytes`
+let field_val = view.size_in_bytes().try_read();
+
+// Accessing the helper method to get the view size
+let view_size = StructView::size_in_bytes(&view);
+```
+
+### Generator Implementation Strategy
+The Rust backend generator will be implemented in Python, integrated directly into the existing Emboss compiler toolchain.
+*   It will reuse the existing Python-based frontend utilities, including the DDL parser, AST representation, and semantic analyzer.
+*   It will also reuse the existing backend framework utilities. Only the C++ generator code and templates will be replaced with Rust counterparts.
+*   Implementing an alternative backend compiler framework is out of scope.
+
+## Alternatives Considered
+
+### 1. Serialization / Deserialization (Copying)
+
+One alternative is to parse the bytes into a standard Rust `struct` on read, and serialize it back to bytes on write.
+
+*   **Pros**: Simple to implement, idiomatic Rust.
+*   **Cons**: Introduces copy overhead (violates Requirement 1). Parsing the entire structure is inefficient if only a single field is needed. Cannot easily handle invalid or partially-written data.
+
+### 2. Direct Layout Mapping (e.g., Zerocopy)
+
+Another alternative is to use safe casting to map a Rust `struct` directly onto a byte slice.
+
+*   **Pros**: Zero-copy, high performance.
+*   **Cons**: Cannot handle complex Emboss features such as:
+    *   Conditional/optional fields.
+    *   Dynamically sized fields.
+    *   Bitfields and non-power-of-two integers (e.g., 24-bit integers).
+    *   Fields with alignment constraints that do not match the CPU.
+
+
+
+## Future Work: Advanced Typestates
+
+To avoid runtime overhead while ensuring safety, we propose using a typestate pattern to track the validation and completeness status of a view at compile time.
+
+A view or writer can be in one of several states:
+*   `UncheckedState`: The initial state (mapping directly to the default fallible approach described above). Fields cannot be accessed directly without runtime bounds checks. Read/write accessors in this state perform dynamic bounds checking on the fly.
+*   `MinimallyCompleteState`: The buffer is known to be at least large enough for the minimum size of the structure.
+*   `CompleteState`: The buffer is large enough for the current dynamic layout of the structure. Always-present fields can be read/written without runtime bounds checks.
+*   `AlwaysCompleteState`: The buffer is large enough for the maximum possible size of the structure.
+*   `OkState`: The structure is both `CompleteState` and logically valid (all `requires` constraints are satisfied).
+*   `AlwaysCompleteOkState`: The structure is both `AlwaysCompleteState` and logically valid.
+
+> [!NOTE]
+> **Dynamic Sub-structures**: A dynamically sized sub-structure cannot inherently inherit the exact completeness typestate from its parent. If a parent is `AlwaysCompleteState`, but it contains a conditionally sized sub-view, the accessed sub-view must have its typestate dynamically downgraded (e.g., down to `MinimallyCompleteState`) upon access, requiring a runtime completeness check on the sub-view's own presence fields to ascend back to `CompleteState`.
+
+State transitions occur when validation functions are called or when fields are mutated (since writing to a field might change the layout or invalidate constraints).
+
+### The Consuming Writer Pattern
+
+Because mutating a layout-affecting field invalidates the bounds of subsequent fields, typestate writers must consume themselves on mutation and return a newly-typed writer representing the transitioned state. 
+
+To use infallible writes, a view is explicitly transitioned into a consuming writer:
+
+```rust
+let mut buf = [0u8; 4];
+let view = BasicType::new(&mut buf)
+    .check_complete()?       // Validates bounds, transitions to CompleteState
+    .into_writer()           // Transitions into a typestate-tracked writer
+    .a().write(0x070A)?      // Consumes and returns the transitioned writer
+    .b().write(0x080B)?
+    .into_view();            // Recovers the base view
+```
+
+```mermaid
+graph TD
+    Unchecked -->|check_complete| Complete
+    MinimallyComplete -->|check_complete| Complete
+    Complete -->|Layout Mutation| MinimallyComplete
+    Complete -->|check_ok| Ok
+    AlwaysComplete -->|check_ok| AlwaysCompleteOk
+    Ok -->|Layout Mutation| MinimallyComplete
+    Ok -->|Constraint Mutation| Complete
+    AlwaysCompleteOk -->|Constraint/Layout Mutation| AlwaysComplete
+```
+#### State Transition Matrix
+The following table defines the target state for a write operation based on the starting state and the field category:
+
+| Initial State | Field Category | Resulting State | Signature Style |
+| :--- | :--- | :--- | :--- |
+| **`AlwaysCompleteState`** | Any Field | `AlwaysCompleteState` | Consuming |
+| **`CompleteState`** | Layout-Neutral | `CompleteState` | Consuming |
+| **`CompleteState`** | Layout-Affecting | `MinimallyCompleteState` | Consuming |
+| **`AlwaysCompleteOkState`** | Layout-Neutral & Constraint-Neutral | `AlwaysCompleteOkState` | Consuming |
+| **`AlwaysCompleteOkState`** | Layout-Affecting / Constraint-Affecting | `AlwaysCompleteState` | Consuming |
+| **`OkState`** | Layout-Neutral & Constraint-Neutral | `OkState` | Consuming |
+| **`OkState`** | Constraint-Affecting | `CompleteState` | Consuming |
+| **`OkState`** | Layout-Affecting | `MinimallyCompleteState` | Consuming |
+
+
+
+For example, mutating a field that affects the presence of another field (e.g., a length field or a conditional flag) will transition the state from `CompleteState` back to `MinimallyCompleteState`.
+
+Consider this structure with a conditional field:
+
+```emboss
+struct Simple:
+  0 [+1]  UInt  a
+  1 [+2]  UInt  b
+  if a == 42:
+    3 [+1]  UInt  c
+```
+
+Writing to `a` may change whether `c` is present. The state tracking ensures that we must re-verify completeness before accessing `c`. The main writer API is always-consuming, meaning writes return the transitioned writer:
+
+```rust
+let writer = Simple::new(&mut buf).check_complete()?.into_writer();
+
+let writer = writer.a().write(42)?; // Changes layout, transitions state: CompleteState -> MinimallyCompleteState
+// writer.c() // Does not compile! Accessor 'c' is not available in MinimallyCompleteState.
+
+let writer = writer.check_complete()?; // Re-verify completeness -> CompleteState
+let writer = writer.c().write(5)?; // Works!
+```
+
+### Mutating Nested Fields
+
+When writing to nested fields (structs or bits), the inner writer must mutate the outer view's buffer. To support this while enforcing typestate safety, the backend uses a nested storage pattern.
+
+Consider the following example structure `SimpleNested` which contains a nested bits block `my_bits`:
+
+```emboss
+struct SimpleNested:
+  0 [+1] bits my_bits:
+    0 [+6] UInt a
+    6 [+2] UInt d
+  1 [+2] UInt b
+  if a == 42:
+    3 [+1] UInt c
+  let a = my_bits.a
+  let d = my_bits.d
+```
+
+In this structure:
+*   `my_bits.a` (aliased as `a`) is **layout-affecting** because it is used in `SimpleNested`'s presence condition for `c`.
+*   `my_bits.d` (aliased as `d`) is **layout-neutral** (invariant) because it does not affect any layout or presence conditions in `SimpleNested`.
+
+#### 1. Nested Storage Wrapper (`NestedStorage`)
+When accessing a nested field's view or writer, the compiler constructs the nested view/writer over a `NestedStorage<Outer>` wrapper.
+*   `NestedStorage` wraps the **outer view or writer (`Outer`)** itself, not another storage.
+*   For views, it borrows the parent (e.g., `NestedStorage<&'a Parent>`), and for writers, it owns it by value (`NestedStorage<Parent::Writer>`) to support the consuming writer pattern.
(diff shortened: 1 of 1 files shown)


doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+
+To avoid runtime overhead while ensuring safety, we propose using a typestate pattern to track the validation and completeness status of a view at compile time.
+
+A view or writer can be in one of several states:
+*   `UncheckedState`: Initial state. Fields cannot be accessed directly without runtime bounds checks. Read/write accessors in this state perform dynamic bounds checking on the fly.
+*   `MinimallyCompleteState`: The buffer is known to be at least large enough for the minimum size of the structure.
+*   `CompleteState`: The buffer is large enough for the current dynamic layout of the structure. Always-present fields can be read/written without runtime bounds checks.
+*   `AlwaysCompleteState`: The buffer is large enough for the maximum possible size of the structure.
+*   `OkState`: The structure is both `CompleteState` and logically valid (all `requires` constraints are satisfied).
+*   `AlwaysCompleteOkState`: The structure is both `AlwaysCompleteState` and logically valid.
```
EricRahm, line 153: It feels like we might want separate states for "big enough" and "valid" (for example I can imagine `CompleteOkState` would be reasonable). We also might consider what states make sense for readers vs writers (they might all make sense, I'm just not sure).

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+*   **Ergonomics**: Compiler error messages involving these types can become long and difficult to read.
+*   **Compile Times**: Deep type recursion can increase Rust compiler monomorphization overhead and compile times for very large schemas.
+This is a conscious trade-off to achieve zero-copy safety and C++ equivalent performance.
+
+#### 3. Field Accessor Signatures (Conditionals and Arrays)
+Field accessors return different types based on whether the field is logically optional, and what the typestate guarantees about the backing storage:
+
+##### Non-Optional Fields (Default)
+*   In **`UncheckedState`** or **`MinimallyCompleteState`**: Accessors return `Result<FieldView, Error>` (or `Result<T, Error>` for primitives). If the storage is too small, they return `Err(Error::OutOfBounds)`.
+*   In **`CompleteState`** or **`AlwaysCompleteState`**: Accessors return `FieldView` (or `T`) directly, as the storage is compile-time guaranteed to be available.
```
EricRahm: I get the intention to be flexible here, but I'm a little worried that we're pushing a fair amount of complexity to developers by having a lot of variability in what presumably are similar interfaces. It might be better to have interfaces that just always return results and let the compiler deal with simplifying things under the hood (if that would even work, I can see an argument for optimization, I'm just not sure if it's necessary or not).

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+  0 [+2]  UInt  a
+  2 [+2]  UInt  b
+```
+
+A reader view can be constructed and accessed as follows:
+
+```rust
+// Example usage of a generated view
+let buf = [0x01, 0x0A, 0x02, 0x0B];
+let view = BasicStruct::new(&buf);
```
EricRahm: We probably need a section of this doc that talks about naming. How do we map emboss names and namespaces to idiomatic rust?

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+*   In **`UncheckedState`** or **`MinimallyCompleteState`**: Accessors return `Result<FieldView, Error>` (or `Result<T, Error>` for primitives). If the storage is too small, they return `Err(Error::OutOfBounds)`.
+*   In **`CompleteState`** or **`AlwaysCompleteState`**: Accessors return `FieldView` (or `T`) directly, as the storage is compile-time guaranteed to be available.
+
+##### Optional (Conditional) Fields
+We distinguish between storage availability errors (`Err(Error::OutOfBounds)`) and logical field absence (`None`):
+*   In **`UncheckedState`** or **`MinimallyCompleteState`**: Accessors return `Result<Option<FieldView>, Error>`.
+    *   If the storage is too small to read the presence condition: returns `Err(Error::OutOfBounds)`.
+    *   If the condition is successfully read and evaluates to `false`: returns `Ok(None)` (even if the storage is too small to hold the field itself).
+    *   If the condition is `true` and storage is valid for the field: returns `Ok(Some(FieldView))`.
+    *   If the condition is `true` but storage is too small for the field: returns `Err(Error::OutOfBounds)`.
```
EricRahm: Do we want to distinguish this somehow from "too small to read condition"?

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+    ```
+
+#### 7. Parameterized Structs
+Emboss supports parameterized structures (e.g., passing runtime configuration down to sub-structures).
+*   Any DDL parameters are stored as fields directly inside the generated structures (such as `BasicStruct<S, ST>`).
+*   When a nested writer is accessed, the parameters are propagated implicitly because the `NestedStorage` holds the parent writer (which contains the parameters) by value.
+
+#### 8. Alignment and LLVM Optimizations
+Unlike C++ which tracks buffer alignment in the template types, the Rust backend delegates alignment optimization to the compiler.
+*   During our prototyping, we investigated the generated assembly and verified that modern compilers (LLVM) are highly effective at optimizing Rust's safe slice operations (`copy_from_slice`, `try_into`) into single, optimal load/store instructions equivalent to hand-written unaligned pointer dereferences.
```
EricRahm: What build targets were we looking at? It'd be nice to understand the behavior for different architectures.

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+
+1.  **Zero-copy**: Accessing fields should not require copying data into intermediate structures.
+2.  **Efficiency**: Generated code should compile to efficient machine code, equivalent to hand-written unsafe pointer manipulation but with safety checks.
+3.  **Safety**: The generated API must be memory-safe. Any use of `unsafe` must be internal to the runtime, minimized, and thoroughly verified.
+4.  **Idiomatic Rust**: The API should feel natural to Rust developers, utilizing features like `Result`, `Option`, and traits.
+5.  **`no_std` Support**: The runtime and generated code must support `no_std` environments for use in embedded systems or kernels.
+
+## Proposed Design
+
+### Views and Storage
```
EricRahm, line 20: It'd be nice to discuss views and storage in more detail here.

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+    3 [+1] UInt c
+  let a = my_bits.a
+  let d = my_bits.d
+```
+
+In this structure:
+*   `my_bits.a` (aliased as `a`) is **layout-affecting** because it is used in `SimpleNested`'s presence condition for `c`.
+*   `my_bits.d` (aliased as `d`) is **layout-neutral** (invariant) because it does not affect any layout or presence conditions in `SimpleNested`.
+
+#### 1. Nested Storage Wrapper (`NestedStorage`)
```
EricRahm, line 249: I'm not sure the right answer here, and this may work well but it's unclear to me: A common pattern is to pass in a portion of a view to another function, so perhaps we can imagine:

```
struct Foo:
  struct Header header: ...
  struct Body body: ...
```

And I had:
```rust
fn process_body(&body: Foo::Body) {...}

let foo = Foo::new(&buf);
// what would passing Foo::Body look like, if at all?
// process_body(foo.body());
```

This might be an anti-pattern we discourage or we might want to consider how to support it. Curious on your thoughts.

EricRahm (commented):
Love this, thanks for drafting it up! I didn't do a super deep review, but just added some conversation starter questions.

doc/design_docs/rust_backend.md
jasongraffius: Actually this was an oversight I left out while editing, I need to split this out into the `try_read` and `read` interfaces. The latter never returns a Result, but is allowed to panic if it can't statically ensure it is safe to access. I completely missed updating this section

doc/design_docs/rust_backend.md
jasongraffius: I can add some clarification here, but this is intended to be fully supported. A struct field has the exact same interface as a struct (the storage would be a slice of the outer storage). You would need to be generic over storage OR provide your own dynamic storage type you use throughout, but emboss wouldn't be opinionated on that

doc/design_docs/rust_backend.md
jasongraffius: The OkState implies Complete, it's not possible in Emboss for a field or struct to be Ok without being Complete (in fact IsOk in C++ essentially starts out with IsComplete() && ...)

As for readers vs writers, completeness matters more to writers and okayness with readers, but it's not completely strict like that

doc/design_docs/rust_backend.md
jasongraffius: Fair enough, I do have a section on naming collisions below (we ensure any valid Emboss name can be generated) but there's more I can explain on the specifics besides just trait dispatch

doc/design_docs/rust_backend.md
jasongraffius: Sure, is there anything specific you're looking for? Also I noticed I left some old verbiage here about storage being a &mut [u8] which isn't the case. That *is* a valid storage but plenty of other things are valid storage too.

jasongraffius (commented):
Thanks, added some comments, I'll update to cover some of the questions here

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+    }
+    ```
+
+#### 7. Parameterized Structs
+Emboss supports parameterized structures (e.g., passing runtime configuration down to sub-structures).
+*   Any DDL parameters are stored as fields directly inside the generated structures (such as `BasicStruct<S, ST>`).
+*   When a nested writer is accessed, the parameters are propagated implicitly because the `NestedStorage` holds the parent writer (which contains the parameters) by value.
+
+#### 8. Alignment and LLVM Optimizations
+Unlike C++ which tracks buffer alignment in the template types, the Rust backend delegates alignment optimization to the compiler.
```
reventlov: First, I agree with the decision to skip over `Aligned` reader/writer types, at least for now: they add complication to end user code that is rarely worth it in practice: the only widely-used target with alignment requirements that I know of, in 2026, is Web Assembly, which has minimal overlap with Emboss's uses.

(There are still a few PowerPC, MicroBlaze, and MIPS processors in use, but they're much more niche.  Plus, the vast majority of uses are totally fine with the slight performance hit of alignment-neutral code even on platforms where it matters.)

(OK, so *technically* there are a few x86 instructions that have alignment requirements, but I think no current compilers emit them, and their unaligned counterparts have nearly the same speed on the last few generations of x86 processors.)

The justification for having `Aligned` *at all* on the C++ side is to hit the marketing benchmark of "equal the performance of packed `struct`s."  (Which also requires using the `Unsafe` methods.)

Second, `rustc` cannot do any better than `clang` or `gcc`, for the same reasons: the compiler cannot always infer the strictest alignment of the base address of a view.  In C++, the `Aligned` view types let the programmer force the compiler to assume a certain *base* alignment, even if that base alignment is not deducible.  (Generally, the base alignment is not deducible when receiving a pointer from some external source, and that is true with `&[u8]` just as much as `char *`.)

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+Emboss currently supports generating C++ code for parsing and validating binary data structures. As Rust becomes more widely used for systems programming and protocol implementations, there is a growing need for native Rust support in Emboss.
+
+This document proposes a design for an Emboss Rust backend. The goal is to generate Rust code that provides the same safety, zero-copy, and efficiency guarantees as the C++ backend, while integrating naturally with Rust's type system and safety features.
+
+## Requirements
+
+A Rust backend for Emboss should satisfy the following requirements:
+
+1.  **Zero-copy**: Accessing fields should not require copying data into intermediate structures.
+2.  **Efficiency**: Generated code should compile to efficient machine code, equivalent to hand-written unsafe pointer manipulation but with safety checks.
```
reventlov: I think I would not make *this* level of efficiency a first-version requirement.  Real-world experience, including on microcontrollers, is that 99% of Emboss-using code does not need absolute maximum performance.  Better to get something that covers most cases first, and only then start taking on the few ultra-performance-critical bits.

Also, in many cases, "absolute maximum performance" means that the end developer has to provide hints (and in Rust, probably go `unsafe`).

See my comments below about the proposed type states and about C++ `Aligned` view types.

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+
+```rust
+let mut buf = [0u8; 4];
+let structure = BasicStruct::new(&mut buf).into_writer()
+    .a().write(0x070A)?
+    .b().write(0x080B)?
+    .into_struct(); // Consumes writer, returns the base struct view
+```
+
+### Compile-Time State Tracking (Typestates)
```
reventlov: My high-level comment about this section is that I would plan to launch with only (the equivalent of) `UncheckedState`, and then worry about adding more states (and more complication to end user code) later, and only in cases where the performance really matters.  Especially since this all falls under the umbrella of "could be done automatically by a sufficiently smart compiler."

doc/design_docs/rust_backend.md
EricRahm: Just wondering about details on how accessing fields might look, optimizations for validated structs, what types of storage would be supported (like do we care about scatter/gather type things, or is this all driven by traits, or...), do we support copy / clone, can there be multiple reader views for the same buffer, etc

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+
+To avoid runtime overhead while ensuring safety, we propose using a typestate pattern to track the validation and completeness status of a view at compile time.
+
+A view or writer can be in one of several states:
+*   `UncheckedState`: Initial state. Fields cannot be accessed directly without runtime bounds checks. Read/write accessors in this state perform dynamic bounds checking on the fly.
+*   `MinimallyCompleteState`: The buffer is known to be at least large enough for the minimum size of the structure.
```
reventlov, line 149: In terms of access patterns, this state seems mostly unnecessary?  It would only allow unchecked access to a somewhat arbitrary set of fields (i.e., those that are static + those that are close enough to the start of a "union" that they're always covered).

That starts getting into "how much do you want to explode the state space?".  You could theoretically get very, very fine-grained.

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+let mut buf = [0u8; 4];
+let structure = BasicStruct::new(&mut buf).into_writer()
+    .a().write(0x070A)?
+    .b().write(0x080B)?
+    .into_struct(); // Consumes writer, returns the base struct view
+```
+
+### Compile-Time State Tracking (Typestates)
+
+To avoid runtime overhead while ensuring safety, we propose using a typestate pattern to track the validation and completeness status of a view at compile time.
```
reventlov, line 145: A `DynamicSizeIsKnownState` might be useful here, as it would let you elide bounds and existence checks when reading fields for size computations.

doc/design_docs/rust_backend.md
jasongraffius: Added

doc/design_docs/rust_backend.md
jasongraffius: AArch64 and x86_64, added language making that clear

doc/design_docs/rust_backend.md
jasongraffius: Added

doc/design_docs/rust_backend.md
jasongraffius: Added a section about this use pattern

doc/design_docs/rust_backend.md
jasongraffius: I don't know that there's much use in doing that, in both cases its "you need more storage or you haven't seen the whole packet yet" so the remedy should be the same.

doc/design_docs/rust_backend.md
jasongraffius: Clarified how this works

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
+# Rust Code Generation Backend for Emboss
```
reventlov, line 1: I may have missed this somewhere, but it would be nice to explain the "escape hatches" you plan to include.  The biggest ones that come to mind are:

1.  Where, how, and in what circumstances can I get to the underlying storage for a view?
2.  Is there a way to read/write beyond the boundary of a specific view (though still within allocated storage)?
3.  If I have some `&[u8]` or `&mut [u8]` or other storage type that I got out of a view, are there any issues with creating a new view over some slice of that storage?  I.e., can I drop out of Emboss, then get back in when there is some logical layer to a structure that is too complex to express in Emboss?

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,583 @@
...
+
+## Requirements
+
+A Rust backend for Emboss should satisfy the following requirements:
+
+1.  **Zero-copy**: Accessing fields should not require copying data into intermediate structures.
+2.  **Efficiency**: Generated code should compile to efficient machine code, equivalent to hand-written unsafe pointer manipulation but with safety checks.
+3.  **Safety**: The generated API must be memory-safe. Any use of `unsafe` must be internal to the runtime, minimized, and thoroughly verified.
+4.  **Idiomatic Rust**: The API should feel natural to Rust developers, utilizing features like `Result`, `Option`, and traits.
+5.  **`no_std` Support**: The runtime and generated code must support `no_std` environments for use in embedded systems or kernels.
```
reventlov: I don't recall whether `no_std` implies "no heap memory allocation," but if not it seems like that would be important to add here.

Also, similar to C++, I would only support `no_std` for "core" functionality, and let things like text serialization/deserialization require `std`.

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,614 @@
...
+
+### Naming Conventions and Conflict Resolution
+
+Emboss's enforced naming conventions align with Rust conventions:
+*   `CamelCase` for struct/enum types maps to Rust `CamelCase` structs/enums.
+*   `snake_case` for fields maps to Rust `snake_case` methods.
+*   `SHOUTY_CASE` for enum values and constants maps to Rust `SHOUTY_CASE` constants.
+
+#### Naming Conflict Resolution
+Emboss structures can have fields with arbitrary names. To avoid naming conflicts between generated field accessors and built-in helper methods (like `size_in_bytes` or `storage`), helper methods are defined on traits (e.g., `emboss_runtime::Struct`).
```
reventlov: Not quite arbitrary: [most Rust keywords are banned](https://github.com/google/emboss/blob/master/compiler/front_end/reserved_words#L916).  However, the [keywords added in the 2018 and 2024 editions](https://doc.rust-lang.org/reference/keywords.html) are *not* banned, so you will need a strategy to handle them.

(You'd have to look up the internal change history to figure out when I actually made that list, but I feel it might have been before Rust 1.0?  It's missing `async`, `await`, and `dyn`, where were reserved but unused sometime before the 2018 edition, and it *has* `alignof`, which seems weird.)

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,614 @@
...
+## Requirements
+
+A Rust backend for Emboss should satisfy the following requirements:
+
+1.  **Zero-copy**: Accessing fields should not require copying data into intermediate structures.
+2.  **Efficiency**: Generated code should compile to efficient machine code, equivalent to hand-written unsafe pointer manipulation but with safety checks.
+3.  **Safety**: The generated API must be memory-safe. Any use of `unsafe` must be internal to the runtime, minimized, and thoroughly verified.
+4.  **Idiomatic Rust**: The API should feel natural to Rust developers, utilizing features like `Result`, `Option`, and traits.
+5.  **`no_std` Support**: The runtime and generated code must support `no_std` environments for use in embedded systems or kernels.
+
```
reventlov, line 17: These are more of policy questions than design questions, but they have some design implications:

What versions of Rust are you planning to support, and what is the plan for how that list changes over time?  I.e., what will you support "at launch," and what is the Rust equivalent of "until 10 years after the release of the following standard" on the C++ side?

My gut feeling is that Rust users are a lot less likely to be tied to ancient toolchains than embedded C++ users, so a much shorter (say, 3-5 year) support window may be fine.  Also, it's fine to only target, say, the Rust 2025 edition at launch, although I don't think I saw anything in here that needs anything that modern.

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,614 @@
...
+### Compile-Time State Tracking (Typestates)
+
+To avoid runtime overhead while ensuring safety, we propose using a typestate pattern to track the validation and completeness status of a view at compile time.
+
+A view or writer can be in one of several states:
+*   `UncheckedState`: Initial state. Fields cannot be accessed directly without runtime bounds checks. Read/write accessors in this state perform dynamic bounds checking on the fly.
+*   `MinimallyCompleteState`: The buffer is known to be at least large enough for the minimum size of the structure.
+*   `CompleteState`: The buffer is large enough for the current dynamic layout of the structure. Always-present fields can be read/written without runtime bounds checks.
+*   `AlwaysCompleteState`: The buffer is large enough for the maximum possible size of the structure.
```
reventlov, line 151: Note that `AlwaysCompleteState` does not imply that sub-structures are actually complete, so any field access for a dynamically-sized field puts you back into `UncheckedState`.  Lots of real-world messages have degenerate encodings where a "payload length" field can have a value that is too short for the actual payload.

doc/design_docs/rust_backend.md
```diff
@@ -0,0 +1,614 @@
...
+    for (i, val) in values.iter().enumerate() {
+      // Writing to an array element is layout-invariant:
+      stable_writer = stable_writer.payload()[i].write(*val)?;
+    }
+    let writer = stable_writer.check_complete()?; // Re-validate once
+    ```
+
+#### 6. Dynamic Writes (State-Stable Structs)
+
+For complex imperative updates where the developer wants to avoid typestate transitions entirely, they can use the writer in `UncheckedState` directly. A struct or writer in a state-stable typestate (such as `UncheckedState` or `MinimallyCompleteState`) is referred to as a **state-stable struct** (or state-stable writer).
```
reventlov: To avoid confusion, I would not call the read-only types structs: they do not hold their data inline the way that an actual `struct` does.  I would either stick with "view" (by analogy with relational database views) and disambiguate with "real-only view" and "read-write view," or consistently use "reader" and "writer."

reventlov (commented):
Overall, I'm excited to see this, especially since I recently started work on a primarily-Rust project.

doc/design_docs/rust_backend.md
jasongraffius: Ah got it, that's a good point, I'll fix up the wording here, but yeah for a first pass we probably want to skip the Aligned stuff for now. FWIW I believe that an AlignedStorage would be implementable by a client as a `Storage` implementation and should work correctly after monomorphization.

doc/design_docs/rust_backend.md
jasongraffius: Fair point, that's true, it's not quite the same level of efficiency. When experimenting, I did try to get to the point that properly used generated code compiled down in assembly to equivalent hand-written implementation, but yeah there's lots of edge cases this would get bogged down on if we needed to actually hit that level across the board. I'll update to clarify!

doc/design_docs/rust_backend.md
jasongraffius: The main reason I came up with the typestate stuff here is because a common paradigm I've seen used for Emboss stuff is "check complete, then write stuff (unchecked or panicking, not error checking), then check okay" not realizing an earlier write could make a later write invalid (either silent UB for the unchecked or crash for panic-checked).

That said, it is technically an optimization, only offering the `try_write` and `try_read` fallible options would be a good starting point. I think I'll reconfigure this to just launch with the `try_` ops without a typestate, and offer the stateful views as a convenience later.

doc/design_docs/rust_backend.md
jasongraffius: Yeah, I figured it was a happy medium between fine-grained explosion and fully coarse, like a common case of access to a known-good fixed header or similar. Honestly this is basically me trying to get Rust to work like a dependently typed system in a way... I'm inclined to keep it as it doesn't add much complexity itself, but you're right there does need to be a fairly arbitrary "where do you stop" line if you add something like this.

doc/design_docs/rust_backend.md
jasongraffius: This was basically my intent with something like MinimallyComplete, but now that I think about it, the dynamic size may require more than just MinimallyComplete (I'm considering a jagged 2D array where each array embeds its length, which is dependent on an outer length)

doc/design_docs/rust_backend.md
jasongraffius: Another reason I liked the typestate is that you can declare your requirements "This function requires a fully complete and validated XyzStruct" means a lot less boilerplate *and* missed/forgotten checks

reventlov: Yeah, just to be clear, I'm not saying "this is a bad idea" or "don't do this;" I'm only suggesting that you defer it until you have something working at all.

doc/design_docs/rust_backend.md
jasongraffius: Yeah it's somewhat orthogonal, but no `alloc` is also my intent, I'll add that.

And yeah, agreed, I'll update this to clarify that "`no_std` support" doesn't mean we can't have things that require `no_std` just there must exist a configuration that supports `no_std`.

doc/design_docs/rust_backend.md
jasongraffius: Ah yes, I actually meant to include that, we want to carve out keywords that aren't in the banlist for `r#keyword` treatment, the only ones you can't use that for are covered by the banlist.

doc/design_docs/rust_backend.md
jasongraffius: I'm trying to think of the implications of this... like a struct containing a sub-struct in a payload, that sub-struct is sized dynamically by some other field, and therefore IsComplete is false. Am I getting that right?

doc/design_docs/rust_backend.md
jasongraffius: By making the storage generic, I was hoping to actually permit the "carries storage with it" style structs. If you have `Simple6ByteStruct::new([u8; 6])`, for instance, that would give you a Simple6ByteStruct<[u8; 6]>. So the difference between a view and owning is whether the storage is a view or owning, and the struct does "own" its storage, just that that storage may itself be a borrow or other referential type.

doc/design_docs/rust_backend.md
reventlov: Doesn't have to be that complicated.  If you have:

```
struct Message:
  # ... fields ...
  8  [+2]  MessageType  message_type (mt)
  10 [+2]  UInt         message_length (len)
  if mt == MessageType.Foo:
    12 [+len]  Foo  foo
  if mt == MessageType.Bar:
    12 [+len]  Bar  bar
  # ... etc ...
```

When `message_length` is `0`, you can have infinite bytes available in `Message`'s storage, but `foo`/`bar`/etc. will never be complete.

jasongraffius: Right right, that's the same scenario, no? `Foo`/`Bar` as the substruct, `len` as the dynamic size by some other field, the inner structs can be incomplete while the outer is complete.

But yeah, that's a good point this should be called out, in the typestate model, dynamically sized sub-structs need to have their state downgraded on access.

doc/design_docs/rust_backend.md
reventlov: I see.  That's a good goal, but I would still avoid the name "struct" for the generic type; let someone make a <code>type Simple6ByteStruct = Simple6Byte*Something*\<[u8; 6]\>;</code> alias for the inline storage version.

doc/design_docs/rust_backend.md
jasongraffius: Yep yep! In agreement here, I'm gonna reconfigure this doc to move this to a future work section.

doc/design_docs/rust_backend.md
reventlov: I may have misread what you said; I thought you were saying that one more level of wrapping was needed.  (Like that `Foo` needed to have a dynamically-sized sub-structure.)

doc/design_docs/rust_backend.md
jasongraffius: > Simple6ByteStruct = Simple6ByteSomething<[u8; 6]>

Sorry, to clarify, I'm not saying "Struct" is appended to the type name, I'm saying that is the type used in Emboss, `struct Simple6ByteStruct`. Maybe it would be clearer if I avoid that in the examples like `struct Simple6BytePacket` would create a Rust type `Simple6BytePacket`. No `Struct` is appended.

reventlov: Ah, that definitely does not come across with the phrases "a struct or writer" and "a state-stable struct (or state-stable writer)."

jasongraffius: Fair point! Let me fix the wording here.

jasongraffius: I'll update some names too, `Struct` -> `StructView` and `into_struct` -> `into_view`.

doc/design_docs/rust_backend.md
jasongraffius: Removed the references to alignment and optimizations

doc/design_docs/rust_backend.md
jasongraffius: Removed this, we definitely aren't trying to get maximally efficient, at least not for now, a future version might be interested in enhancements here

doc/design_docs/rust_backend.md
jasongraffius: Added not needing alloc as a requirement, and clarified that we can have features that require std as long as a `no_std` configuration is supported

doc/design_docs/rust_backend.md
jasongraffius: Added

doc/design_docs/rust_backend.md
jasongraffius: Added a section discussing this. Basically I'm proposing we land the initial backend as "experimental" without much support guarantees, notifying the emboss mailing list for major breaking changes, and deferring specific decisions to stabilization. I do propose that we target 2024 edition only during the experimental phase, but I also agree I don't think we use anything that would preclude support for earlier editions

doc/design_docs/rust_backend.md
jasongraffius: Clarified that sub-views cannot always inherit state of outer view

doc/design_docs/rust_backend.md
jasongraffius: I've moved typestates to future work, we can hash out the specifics in a separate typestate design

doc/design_docs/rust_backend.md
jasongraffius: I've moved typestates to future work, we can hash out the specifics in a separate typestate design

doc/design_docs/rust_backend.md
jasongraffius: Added an escape hatches section. I mentioned getting the storage, but it was kind of glossed over. You can borrow storage with `.storage()` and take the storage with `into_storage()`.

PR #271, opened 2026-07-19
Fix unsound Ok() switch guard by demoting all-residual conditional discriminants

AaronWebster:
## Problem

The optimized `Ok()` switch (introduced in #253, extended in #256/#257) reads its
discriminant once and emits, as a safety net:

```cpp
const auto emboss_reserved_switch_discrim = <discriminant>;
if (!emboss_reserved_switch_discrim.Known()) return false;
switch (emboss_reserved_switch_discrim.ValueOrDefault()) { ... }
```

When the discriminant is itself a **conditionally-present field** read from a buffer too
short to contain it, it reads as `Unknown`, so this guard makes `Ok()` return `false` —
even though the message is valid, because the fields gated on the (absent) discriminant
simply don't exist. This is latent on `master`: no existing test schema produces the
shape.

Same bug that #266 fixes, reached by a simpler route.

## Fix

Instead of #266's dedicated `ok_method_switch_block_guarded` template, add one rule to
the demotion pass in `_generate_optimized_ok_method_body`: a switch group that is **not
provably-known** *and* has **no bare arm** (every entry carries a residual) is demoted to
per-field `has_X()` checks — the existing `ok_method_test` path — instead of being
emitted as a guarded `switch`. After this, `_emit_switch_block` only ever sees
provably-known or bare-arm-guarded forms, so the unsound guarded-switch template is
deleted outright.

Soundness is observationally identical to #266's guarded fallback: with the discriminant
`Unknown`, a statically-false residual makes `has_X()` Known-false (field absent → valid,
the exact case #266 fixed), otherwise `has_X()` is `Unknown` → `Ok()` returns false; with
the discriminant `Known` it computes the full `has_X()`, same as a dispatched arm minus
the pointless `switch`.

## Verification

- Proven **byte-identical** generated code on every existing schema; the golden diff is
  confined to `testdata/golden_cpp/condition.emb.h`.
- Keeps #266's regression tests (`testdata/condition.emb` +
  `compiler/back_end/cpp/testcode/condition_test.cc`); assertions unchanged, comments
  reworded per review.
- Full `bazel test //...` green (206/206).
- **Size: ~0 delta on all targets** (MicroBlaze LE / Vitis authoritative) — this is a
  correctness fix, not a size optimization.

## Supersedes

Supersedes **#266** (same bug, simpler fix; #266's regression tests are carried over here).

### DIFF
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -1679,14 +1679,33 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
             groups[key]["fields"].append(field)
             field_group_key[id(field)] = key
 
-    # Demote switch groups that wouldn't actually benefit from a switch.
-    # A switch's overhead — the temporary discriminant variable, the
-    # Known() guard, the scope braces — is only worthwhile if the switch
-    # dedupes either at least two distinct case labels (sharing one
-    # subexpression read) or at least two distinct fields (sharing one
-    # discriminant evaluation). When a group has just one entry total
-    # (one field, one bare case), the unoptimized has_${field}()-based
-    # check is strictly smaller.
+    # Demote switch groups back to per-field checks in two cases.
+    #
+    # (1) Size: a switch's overhead — the temporary discriminant variable,
+    #     the Known() guard, the scope braces — is only worthwhile if the
+    #     switch dedupes either at least two distinct case labels (sharing
+    #     one subexpression read) or at least two distinct fields (sharing
+    #     one discriminant evaluation). When a group has just one entry
+    #     total (one field, one bare case), the unoptimized has_${field}()-
+    #     based check is strictly smaller.
+    #
+    # (2) Correctness: the switch reads its discriminant once and guards it
+    #     with `if (!discrim.Known()) return false;`. That guard is only
+    #     sound when the discriminant is provably present OR the group has
+    #     at least one *bare* arm (existence condition exactly `discrim ==
+    #     K`, no residual). With a bare arm, an Unknown (out-of-bounds)
+    #     discriminant leaves that field's has_X() Unknown, so the
+    #     unoptimized Ok() bails for the same reason the guard does. But
+    #     when the discriminant is itself a conditionally-present field and
+    #     *every* arm carries a residual, a valid message can have an
+    #     Unknown discriminant while every arm's residual is statically
+    #     false — making every has_X() Known-false (absent). The blanket
+    #     guard would wrongly reject it. Demote that shape to per-field
+    #     has_${field}() checks, which handle an Unknown discriminant
+    #     correctly (an Unknown discriminant can never make has_X()
+    #     Known-true, so a Known-false residual leaves the field simply
+    #     absent). This is latent on the current test corpus — no existing
+    #     schema produces the shape — so no existing golden changes.
     for key in ordered_keys:
         group = groups[key]
         if group["type"] != "switch":
@@ -1696,6 +1715,16 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
             for case_entry in group["cases_by_label"].values()
         )
         if total_entries < 2:
+            group["type"] = "demoted_to_if"
+            continue
+        has_bare_arm = any(
+            not has_residual
+            for case_entry in group["cases_by_label"].values()
+            for (_field, has_residual) in case_entry["entries"]
+        )
+        if not has_bare_arm and not _is_discriminant_provably_known(
+            group["discrim_expr"], fields
+        ):
             group["type"] = "demoted_to_if"
 
     blocks = []
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -1679,7 +1679,7 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
             groups[key]["fields"].append(field)
             field_group_key[id(field)] = key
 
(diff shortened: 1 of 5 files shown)


testdata/condition.emb
```diff
@@ -229,3 +229,99 @@ struct ConditionalOnFlag:
...
   if enabled:
     1 [+1]  UInt  value
+
+
+struct ResidualConditionalDiscriminant:
+  # `tag` is itself conditional, and is the switch discriminant for `a`/`b`
+  # (it is the first conjunct of their existence conditions).  Every arm
+  # carries a residual (`outer == 1`), so when `outer != 1` the message is
+  # valid even though `tag` is absent -- and on a buffer too short to hold
+  # `tag`, reading the discriminant is Unknown.  The optimized Ok() must not
```
robrussell: ~~The optimized~~ Ok() must not reject such a message.

testdata/condition.emb
```diff
@@ -229,3 +229,99 @@ struct ConditionalOnFlag:
...
+
+  if tag == 0 && outer == 1:
+    2 [+1]  UInt  a
+
+  if tag == 1 && outer == 1:
+    2 [+1]  UInt  b
+
+
+struct BareConditionalDiscriminant:
+  # Like above, but `a`/`b` reference the conditional discriminant `tag`
```
robrussell: "Like `ResidualConditionalDiscriminant`, but..."

testdata/condition.emb
```diff
@@ -229,3 +229,99 @@ struct ConditionalOnFlag:
...
+
+
+struct DominatedBareDiscriminant:
+  # Bare arms on a conditional discriminant, plus an always-present `tail`
+  # field that dominates the structure size.  When `outer != 1`, `tag` is
+  # absent and has_a()/has_b() are Unknown, but `tail` keeps the size Known so
+  # IsComplete() is true.  Ok() must still be false: a/b's existence is
+  # indeterminate, and the discriminant Known() guard -- not IsComplete() --
+  # is what enforces that.  (A `if (has_tag().ValueOrDefault())` wrapper would
+  # wrongly accept this message.)
```
robrussell: This last phrase is unclear
"(A `if (has_tag().ValueOrDefault())` wrapper would wrongly accept this message.)"

Does that reflect the state before this PR or after? Is that intended behaviour? Some of the explanations in the comments for these tests might be helpful in the language docs (either as clarifications or examples).

testdata/condition.emb
```diff
@@ -229,3 +229,99 @@ struct ConditionalOnFlag:
...
+
+  if (tag == 0 || tag == 1) && outer == 1:
+    2 [+1]  UInt  a
+
+  if (tag == 2 || tag == 3) && outer == 1:
+    2 [+1]  UInt  b
+
+
+struct SingleEntryConditionalDiscriminant:
+  # A single residual arm on a conditional discriminant.  The switch group has
```
robrussell: This implementation detail doesn't belong here:
"The switch group has one entry, so it is demoted to a has_X() check instead of a switch; no discriminant guard is emitted, so the bug cannot arise here."

compiler/back_end/cpp/testcode/condition_test.cc
```diff
@@ -300,6 +300,217 @@ TEST(Conditional, StructWithNestedConditionIsOkWhenOuterConditionExists) {
...
+  // outer == 1, tag == 0: `a` present and readable; `tail` present.
+  ::std::uint8_t buffer[4] = {1, 0, 5, 9};
+  auto writer = DominatedBareDiscriminantWriter(buffer, sizeof buffer);
+  EXPECT_TRUE(writer.Ok());
+  EXPECT_TRUE(writer.has_a().Value());
+  EXPECT_EQ(5, writer.a().Read());
+  EXPECT_EQ(9, writer.tail().Read());
+}
+
+// Disjunction arms (#256 matcher) on a conditional discriminant, all carrying a
```
robrussell: `#256` won't get linked except on github. Same below.

compiler/back_end/cpp/header_generator.py
```diff
@@ -1679,14 +1679,33 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
...
+    #     unoptimized Ok() bails for the same reason the guard does. But
+    #     when the discriminant is itself a conditionally-present field and
+    #     *every* arm carries a residual, a valid message can have an
+    #     Unknown discriminant while every arm's residual is statically
+    #     false — making every has_X() Known-false (absent). The blanket
+    #     guard would wrongly reject it. Demote that shape to per-field
+    #     has_${field}() checks, which handle an Unknown discriminant
+    #     correctly (an Unknown discriminant can never make has_X()
+    #     Known-true, so a Known-false residual leaves the field simply
+    #     absent). This is latent on the current test corpus — no existing
```
robrussell: "This is latent on the current test corpus — no existing schema produces the shape — so no existing golden changes." will not be true after this PR. Line 1682 seems to refer to the current state of the codebase as well.

compiler/back_end/cpp/header_generator.py
```diff
@@ -1697,6 +1716,16 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
         )
         if total_entries < 2:
             group["type"] = "demoted_to_if"
+            continue
+        has_bare_arm = any(
+            not has_residual
+            for case_entry in group["cases_by_label"].values()
+            for (_field, has_residual) in case_entry["entries"]
```
robrussell: `for (_, has_residual) in case_entry["entries"]`

robrussell (approved):

PR #272, opened 2026-07-19
Eliminate redundant residual re-read in optimized Ok() switch arms

AaronWebster:
Based on #271 (the guard-demote correctness precursor); review/merge that first.

## What

In the optimized `Ok()` switch, each residual-guarded arm previously re-read and
re-compared the discriminant that the `switch` had already decoded, then recomputed
`has_X()`. This gates each residual arm on the **residual alone**:

```cpp
if (!(<R>).Known()) return false;
if ((<R>).ValueOrDefault() && !X().Ok()) return false;
```

dropping the redundant discriminant re-read. For more than one conjunct the residual is
rendered once as a synthesized boolean-`AND` expression (shared subexpressions). The gate
is required, not merely an optimization: `X()` internally re-checks `has_X()` and returns
a null view when the field is absent, so an unguarded `X().Ok()` would wrongly fail.

Soundness rests on the #271 invariant: every switch reaching `_emit_switch_block` has the
discriminant `Known` inside `case K:` (provably, or via the emitted `Known()` guard), so
`has_X() ⟺ residual`.

## Measurement

Adds a residual-heavy `OuterGuardedUnion` schema to `testdata/many_conditionals.emb`
(always-present `outer`, ~100 arms guarded on `tag == K && outer == 1`) so Lever A
actually fires and the committed harness (#270) can measure it. The flat
`LargeConditionals` union is all bare arms and is the unaffected control.

`OuterGuardedUnion::Ok()` reader symbol @ `-Os`, vs the #271 baseline:

| Target | before → after | Δ |
|---|---|---|
| MicroBlaze LE (Vitis, **authoritative**) | 16748 → 13620 | **−18.7%** |
| MicroBlaze BE | 13960 → 10964 | −21.5% |
| ARM Cortex-M4 | 9296 → 7716 | −17.0% |
| x86-64 (g++) | — | −9.5% |
| x86-64 (clang) | — | −50.7% |

Retired instructions (OuterGuardedUnion hot loop): MB-BE −3.2%, MB-LE −1.6%, ARM −1.1%
(small — only the one active arm runs per call).

**Control `LargeConditionals`: byte-identical codegen, 0 symbol delta on every target.**
(The plan predicted x86 size unchanged; Lever A in fact helps x86 too.)

## Verification

- Golden diff confined to `condition.emb.h` + `many_conditionals.emb.h`; all pre-existing
  structs byte-identical.
- Full `bazel test //...` green (206/206), including `condition_test`'s #266 oracle.

### DIFF
--- a/compiler/back_end/cpp/header_generator.py
+++ b/compiler/back_end/cpp/header_generator.py
@@ -1663,7 +1663,10 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
                 # case label twice (`tag == 0 || tag == 0`); after #9 the
                 # simplifier should normally prevent this from reaching us.
                 if not any(e[0] is field for e in case_entry["entries"]):
-                    case_entry["entries"].append((field, bool(residual)))
+                    # Keep the residual conjunct list itself (not just a bool):
+                    # a residual arm gates its Ok() check on the residual alone
+                    # at emit time (Lever A), so the concrete IR must survive.
+                    case_entry["entries"].append((field, residual))
             field_group_key[id(field)] = key
         else:
             cond_res = _render_expression(
@@ -1718,9 +1721,9 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
             group["type"] = "demoted_to_if"
             continue
         has_bare_arm = any(
-            not has_residual
+            not residual
             for case_entry in group["cases_by_label"].values()
-            for (_field, has_residual) in case_entry["entries"]
+            for (_field, residual) in case_entry["entries"]
         )
         if not has_bare_arm and not _is_discriminant_provably_known(
             group["discrim_expr"], fields
@@ -1740,7 +1743,7 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
             group["known_check_required"] = not _is_discriminant_provably_known(
                 group["discrim_expr"], fields
             )
-            blocks.append(_emit_switch_block(group))
+            blocks.append(_emit_switch_block(group, ir, subexpressions))
         elif group["type"] == "demoted_to_if":
             for field in group["encounter_order"]:
                 blocks.append(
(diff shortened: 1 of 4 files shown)


compiler/back_end/cpp/header_generator.py
```diff
@@ -1663,7 +1663,10 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
                 # case label twice (`tag == 0 || tag == 0`); after #9 the
                 # simplifier should normally prevent this from reaching us.
                 if not any(e[0] is field for e in case_entry["entries"]):
-                    case_entry["entries"].append((field, bool(residual)))
+                    # Keep the residual conjunct list itself (not just a bool):
+                    # a residual arm gates its Ok() check on the residual alone
+                    # at emit time (Lever A), so the concrete IR must survive.
```
robrussell, line 1668: This comment seems unnecessary.

compiler/back_end/cpp/header_generator.py
```diff
@@ -1718,9 +1721,9 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
             group["type"] = "demoted_to_if"
             continue
         has_bare_arm = any(
-            not has_residual
+            not residual
             for case_entry in group["cases_by_label"].values()
-            for (_field, has_residual) in case_entry["entries"]
+            for (_field, residual) in case_entry["entries"]
```
robrussell, line 1726: `for (_, residual) in case_entry["entries"]`

compiler/back_end/cpp/header_generator.py
```diff
@@ -1761,37 +1764,63 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
...
-    then trusted to fold the now-trivially-true discriminant comparison
-    inside the has_${field}() call (it's inlined and the case label has
-    pinned the discriminant value).
+    Each entry is `(field, residual)` where `residual` is the list of
+    predicate conjuncts (IR sub-expressions) that the field's existence
+    condition carries *beyond* the discriminant equality that routed it to
+    this case. When the list is empty the arm is bare and the case body is
+    a single direct Ok() check.
+
+    When there is a residual, we gate the Ok() check on the residual *alone*
```
robrussell, line 1776: "When the list of residuals is not empty..."

compiler/back_end/cpp/header_generator.py
```diff
@@ -1761,37 +1764,63 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
...
-    Each entry is `(field, has_residual)` where `has_residual` indicates
-    whether the field's existence condition has predicate conjuncts beyond
-    the discriminant equality. When there is no residual the case body is
-    a single direct Ok() check; when there is a residual the body falls
-    back to the has_${field}() accessor, which encapsulates the full
-    existence check including the residual conjuncts. The C++ compiler is
-    then trusted to fold the now-trivially-true discriminant comparison
-    inside the has_${field}() call (it's inlined and the case label has
-    pinned the discriminant value).
+    Each entry is `(field, residual)` where `residual` is the list of
```
robrussell, line 1770: `residuals`

compiler/back_end/cpp/header_generator.py
```diff
@@ -1761,37 +1764,63 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
...
+    The gate is required, not just an optimization: `${field}()` re-checks
+    `has_${field}()` internally and returns a null view (whose Ok() is false)
+    when the field is absent, so an unguarded `${field}().Ok()` would wrongly
+    fail whenever the residual is false. We first bail if the residual is not
+    Known (e.g. an out-of-bounds read), matching the has_${field}()-based
+    check it replaces.
     """
     parts = []
-    for field, has_residual in entries:
+    for field, residual in entries:
```
robrussell, line 1792: `residuals`

compiler/back_end/cpp/header_generator.py
```diff
@@ -1761,37 +1764,63 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
...
+    emitted `if (!discrim.Known()) return false;` guard), so within the arm
+    the discriminant equality is already established and `has_${field}() <=>
+    residual`. Gating on the residual therefore avoids re-reading and
+    re-comparing the discriminant that `has_${field}()` would recompute.
+
+    The gate is required, not just an optimization: `${field}()` re-checks
+    `has_${field}()` internally and returns a null view (whose Ok() is false)
+    when the field is absent, so an unguarded `${field}().Ok()` would wrongly
+    fail whenever the residual is false. We first bail if the residual is not
+    Known (e.g. an out-of-bounds read), matching the has_${field}()-based
```
robrussell, line 1788: "matching the has_${field}()-based check it replaces." refers to code that no longer exists after this PR.

The explanation is somewhat useful but overly verbose and includes a lot of implementation details about this function and others which may not hold true in future revisions.

compiler/back_end/cpp/header_generator.py
```diff
@@ -1761,37 +1764,63 @@ def _generate_optimized_ok_method_body(fields, ir, subexpressions):
...
+    check it replaces.
     """
     parts = []
-    for field, has_residual in entries:
+    for field, residual in entries:
         name = _cpp_field_name(field.name.name.text)
-        if has_residual:
+        if residual:
```
robrussell, line 1794: These branches can be simplified, something like:

```
for field, residuals in entries:
    name = _cpp_field_name(field.name.name.text)
    if len(residuals) == 0:
        parts.append("          if (!{}().Ok()) return false;\n".format(name))
        continue
    if len(residuals) == 1:
        residual = residuals[0]
    else:
        residual = ir_data.Expression(
            function=ir_data.Function(
                function=ir_data.FunctionMapping.AND,
                args=residuals,
            ),
            type=ir_data.ExpressionType(boolean=ir_data.BooleanType()),
        )
    rendered = _render_expression(
        residual, ir, subexpressions=subexpressions
    ).rendered
    parts.append(
        "          if (!({0}).Known()) return false;\n          if (({1}).ValueOrDefault() && !{2}().Ok()) return false;\n"
            .format(rendered, rendered, name
        )
    )
```

robrussell (approved):

PR #276, opened 2026-08-03
rust: Ignore Cargo build artifacts in .gitignore

yi-json:
Add `target` and `Cargo.lock` to `.gitignore` so `cargo test` runs inside `runtime/experimental/rust` do not pollute git status

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@ __pycache__
 # IDEs
 .vscode
 
+
+# Rust / Cargo
+**/rust/Cargo.lock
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -9,3 +9,7 @@ __pycache__
 # IDEs
 .vscode
 
+
+# Rust / Cargo
+Cargo.lock
+target/
```
jasongraffius: Can we restrict this to being under a subdirectory named `rust/`?

